PR #7, opened 2022-02-20
Incorporate Change Split Logic

henryyuanheng-wang:
Initial review of change split logic. Passing existing test_txbuilder cases. Will add additional test cases after discussion.


pycardano/txbuilder.py
```diff
@@ -199,6 +231,102 @@ def _add_change_and_fee(
...
+
+            if len(updated_amount.to_cbor("bytes")) > max_val_size:
+                output.amount = old_amount
+                break
+
+            multi_asset_arr.append(output.amount.multi_asset)
+
+        return multi_asset_arr
+
+    def _fees_for_output(self):
```
cffls: Looks like this is not used anywhere.

cffls:
Thanks Henry for contributing such an important feature. Integration tests are passed, which is nice! Builds failed because some code are not well formatted. You can format the code by running `make format` and verify the format by `make qa`. Also, all existing tests probably are using very small set of assets, so there isn't a case where the assets are actually split into multiple outputs. Could you please add an explicit test for that? Thank you!

henryyuanheng-wang:
Yes absolutely Jerry! Wanted to see whether the minting integration test pass. Will format the code and add the test cases to cover the feature of the split functions. thanks!

test/pycardano/test_txbuilder.py
```diff
@@ -200,6 +204,82 @@ def test_tx_builder_mint_multi_asset(chain_context):
     assert expected == tx_body.to_primitive()
 
 
+def test_tx_add_change_split_nfts(chain_context):
+    # Set the max value size to be very small for testing purpose
```
cffls, line 208: Love this smart approach!

cffls:
Looks good to me. I will squash the commits and merge it to main branch. Thank you so much for making this happen!

henryyuanheng-wang:
Thanks Jerry for reviewing and approving the request!

PR #12, opened 2022-02-22
returning pending tx id from submit_tx

34r7h:


cffls:
Hi,

Thanks for the pull request. The transaction ID is deterministic and could also be retrieve from [transaction body](https://github.com/cffls/pycardano/blob/main/pycardano/transaction.py#L318-L320) and [transaction](https://github.com/cffls/pycardano/blob/main/pycardano/transaction.py#L333-L335). Do you think they are sufficient for your use case? I am conservative about using the return value simply for transaction ID because there might be something else that could be returned from backend services in the future.

34r7h:
I had a feeling it was deterministic but didn't know about the method.
Thanks, I'm satisfied however perhaps it's worthwhile to return the id
instead of no return, keeping the example as is to address the concern if
blockfrost decides to change their API. All good whichever you decide.
Thanks again!

cffls:
Thanks @34r7h . Not all submission API will return transaction ID. For example, Ogmios will simply return a "SubmitSuccess" string (see [here](https://ogmios.dev/mini-protocols/local-tx-submission/)). For better consistency, I will not merge this change at the moment. Anyway, this is a good discussion. Thank you for contribution, and I am looking forwards to seeing more in the future!

34r7h:
Ah, ok that makes sense. Thanks for explaining. Force be with!

PR #27, opened 2022-04-20
Fee Estimation Improvement

henryyuanheng-wang:
Replaced max fee estimation with the more accurate fee estimation function. Created additional unit tests to check:

- When input utxo value is small but greater than requested and respect min ada required for change
- When input utxo value is small and remainder value is smaller than min ada required for change
- When input utxo value is small and remainder value is smaller than min ada required for change, but there are additional utxos in the address that can be selected

All unit tests pass and the integration test passes.


test/pycardano/test_txbuilder.py
```diff
@@ -189,6 +190,92 @@ def test_tx_too_big_exception(chain_context):
...
+    tx_in1 = TransactionInput.from_primitive([b"1" * 32, 3])
+    tx_out1 = TransactionOutput.from_primitive([sender, 4000000])
+    utxo1 = UTxO(tx_in1, tx_out1)
+
+    tx_builder.add_input(utxo1).add_input_address(sender_address).add_output(
+        TransactionOutput.from_primitive([sender, 3000000])
+    )
+
+    # Balance is smaller than minimum ada required in change
+    # No more UTxO is available, throwing UTxO selection exception
```
cffls: This comment is irrelevant to this test right?

pycardano/txbuilder.py
```diff
@@ -111,10 +111,8 @@ class TransactionBuilder:
 
     def add_input(self, utxo: UTxO) -> TransactionBuilder:
         """Add a specific UTxO to transaction's inputs.
-
```
cffls: Could you keep these empty lines? They are helpful to readability.
Tip: `make format` might be able to restore them automatically. 😉

pycardano/txbuilder.py
```diff
@@ -601,6 +593,20 @@ def build(self, change_address: Optional[Address] = None) -> TransactionBody:
...
+        plutus_execution_units = ExecutionUnits(0, 0)
+        for redeemer in self.redeemers:
+            plutus_execution_units += redeemer.ex_units
+
+        min_fee = fee(
+            self.context,
+            len(self._build_full_fake_tx().to_cbor("bytes")),
+            plutus_execution_units.steps,
+            plutus_execution_units.mem,
+        )
```
cffls: This pattern is used in multiple places (also in line 329 above). Could you wrap it into a private function? So it could be used like this:
```
min_fee = self._estimate_fee()
```

cffls:
Overall looks really good! Just left a few minor comments. Thanks Henry for the contribution!

pycardano/txbuilder.py
henryyuanheng-wang: Makes sense!

henryyuanheng-wang:
Thanks Jerry for the comments! Made the corresponding changes as suggested.

cffls:
Awesome, LGTM. Thanks Henry!

PR #38, opened 2022-05-15
Merge change into already existing output when possible

astrojarred:
The proposed changes do the following:

- If the change address is already included in one of the outputs, the change (minus the fee) will be added to _last_ output directed towards the change address.
- If no outputs are provided but a change address is provided, the entirety of all input UTxOs will be sent to the change address
- If you want to force the change address into a separate output, create a dummy output like so as the _last_ output added `builder.add_output(TransactionOutput(change_address, 0))`. This will force the change into a new output UTxO even if the change address is already included among the outputs.

FYI I also added a method to `transaction.Output` called `copy()` which essentially deep copies itself. I found it useful for estimating the fees but I'm definitely open to other ways of doing this.

@cffls I updated all the relevant tests in `test_txbuilder.py` to reflect the new fees and outputs. If you have the chance to take a look and make sure I didn't fundamentally alter the purpose of any of them...
Especially `test_tx_add_change_split_nfts` as I couldn't figure out exactly what it was trying to achieve.


pycardano/transaction.py
```diff
@@ -277,6 +277,14 @@ def lovelace(self) -> int:
...
             return self.amount.coin
 
+    def copy(
+        self, dummy_amount: Optional[Union[int, Value]] = None
+    ) -> TransactionOutput:
+        """Return a copy of itself."""
+        return TransactionOutput(
+            self.address, dummy_amount if dummy_amount else self.amount, self.datum_hash
+        )
+
```
cffls: Do you think it will be better to implement `__copy__` in the TransactionOutput class, use `copy(tx_output)`, and reset amount if needed in the caller? It would be in consistent with the copy implementation of some other classes in this package.

pycardano/txbuilder.py
```diff
@@ -350,7 +366,16 @@ def _add_change_and_fee(
             changes = self._calc_change(
                 self.fee, self.inputs, self.outputs, change_address, precise_fee=True
             )
-            self._outputs += changes
+
+            if change_output_indices:
```
cffls: Could we provide a parameter (e.g. `merge_change`) in `build()` for users to control this behavior? Change merging should be disabled by default, and only enabled when users turn it on. This is because merging change into one of the outputs can have undesired/unexpected results, e.g. if some tx outputs need to have exact value amount, adding change randomly to one of them will confuse users.

pycardano/txbuilder.py
```diff
@@ -350,7 +366,16 @@ def _add_change_and_fee(
...
+                # Add the leftover change to the TransactionOutput containing the change address
+                self._outputs[change_output_indices[-1]].amount = (
+                    changes[0].amount + self._outputs[change_output_indices[-1]].amount
```
cffls: Change should be split into multiple UTxOs when one single UTxO can't hold everything because of max value size limit, see [this](https://github.com/input-output-hk/cardano-ledger/blob/master/doc/explanations/min-utxo-alonzo.rst#constraint-on-token-bundle-size). This is why changes can sometimes have more than one UTxO. When that happens, I think we should skip change merge and simply add them to output list, because we have to create additional outputs in that case anyway.

test/pycardano/test_txbuilder.py
```diff
@@ -52,12 +52,10 @@ def test_tx_builder(chain_context):
...
         1: [
-            # First output
-            [sender_address.to_primitive(), 500000],
-            # Second output as change
-            [sender_address.to_primitive(), 4334587],
+            # Change automatically added to sender output
+            [sender_address.to_primitive(), 4836215],
         ],
-        2: 165413,
+        2: 163785,
```
cffls: Like mentioned before, if we disable change merging by default, we won't need to change these test cases.☺️

cffls (commented):
Thanks @astrojarred for working on this great feature! I am so happy to see it. The code looks very good already. I added some comments to the code. The only one major thing we might want to change in this PR is to disable change merging by default to avoid bringing unexpected or surprising behaviors to users. See more details in my inline comments.

Btw, integration tests failed. It seems to be caused by adding a `Value` type to an `int` type. Check more details [here](https://github.com/cffls/pycardano/runs/6444512832?check_suite_focus=true).

astrojarred:
Thanks for the really insightful comments, @cffls. I'll go ahead and get to work on this and let you know if I have any followup questions.

pycardano/transaction.py
astrojarred: Great idea! I didn't even know there was a `__copy__` method from deepcopy. I will implement this similarly to how you've done it in `serialization.DictCBORSerializable`.

pycardano/txbuilder.py
astrojarred: Thanks for taking the time to explain this. I guess this should work simply by checking the length of `changes`?

astrojarred:
> Btw, integration tests failed. It seems to be caused by adding a Value type to an int type.

I will go ahead and fix this, but if preferred I can submit a smaller PR tonight which enforce that all `TransactionOutput.amount` are of type Value (as briefly discussed on Discord).

Briefly trying it out I saw that all tests pass except for one (`test_not_enough_input_amount`) so I hope it might not be too complex to include already.

cffls:
> but if preferred I can submit a smaller PR tonight which enforce that all TransactionOutput.amount are of type Value (as briefly discussed on Discord).

Thanks for the update @astrojarred ! Feel free to submit another PR for it. I can do that later too if you are too busy in fixing this PR. Really appreciate your helps!

astrojarred:
Oh dang, thought I linted this alright. Will merge in the new PR and push again anyway :)

cffls:
Thank you @astrojarred! All tests are passing now, that is awesome! I will try adding some unit tests before merging.

astrojarred:
That's great, I'm so glad! I was also just thinking now about tests. One I thought of is:

- sending an entire UTxO in a single output (fee gets subtracted from UTxO) by adding no outputs and just specifying a change address

PR #39, opened 2022-05-18
Enable UTxO query with Kupo

henryyuanheng-wang:
- Implemented UTxO query alternatives to Ogmios using Kupo (note: currently, Kupo returns all UTxOs for a queried address, regardless of whether they are spent or not. Based on discussion [here](https://github.com/CardanoSolutions/kupo/discussions/19), feature for adding spent flag is planned for Kupo. This will likely happen before next Ogmios deprecates UTxO query support so our utxo query wouldn't be impacted). 

- Updated integration test for querying utxo with Kupo. 

- Added Kupo to docker compose file


integration-test/run_tests.sh
```diff
@@ -11,7 +11,7 @@ poetry install
 ./bootstrap.sh local-alonzo
 
 # Cleanup containers and volumes in case there is any running
-docker-compose down --volume
+docker-compose down --volumes
```
cffls: Nice catch!

pycardano/backend/ogmios.py
```diff
@@ -25,13 +26,24 @@
...
 
 class OgmiosChainContext(ChainContext):
-    def __init__(self, ws_url: str, network: Network, compact_result=True):
+    def __init__(
+        self,
+        ws_url: str,
+        network: Network,
+        compact_result=True,
+        support_kupo=False,
+        http_url=None,
```
cffls: We probably don't need a separate parameter for `support_kupo`. Whenever a kupo url is provided by users, we know kupo should be enabled. Also, what do you think about renaming `http_url` to `kupo_url`?

pycardano/backend/ogmios.py
```diff
@@ -25,13 +26,24 @@
...
         self._ws_url = ws_url
         self._network = network
         self._service_name = "ogmios.v1:compact" if compact_result else "ogmios"
+        self._support_kupo = support_kupo
+        self._http_url = http_url if support_kupo else None
         self._last_known_block_slot = 0
         self._genesis_param = None
         self._protocol_param = None
+        if self._support_kupo and not self._http_url:
+            raise Exception("Cannot find http url to request from Kupo.")
```
cffls: These could be simplified if we only have `kupo_url`.

pycardano/backend/ogmios.py
```diff
@@ -203,6 +267,22 @@ def utxos(self, address: str) -> List[UTxO]:
 
         return utxos
 
+    def utxos(self, address: str, use_kupo=False) -> List[UTxO]:
```
cffls: Let's limit the control of kupo to the constructor of chain context - whenever kupo_url is provided, we will use kupo.

cffls (commented):
Thanks Henry @henryyuanheng-wang, This is awesome! The PR looks really good already. I just had some minor comments.

pycardano/backend/ogmios.py
henryyuanheng-wang: Thanks for reviewing Jerry! All the comments make sense. Question for discussion on this one: right now, the UTxO querying with Kupo will return all UTxOs associated with an address (instead of Unspent UTxOs from Ogmios querying). This will impact how we choose UTxOs from an address when requested amount is greater than provided amount. If utxo query goes through Kupo when kupo_url is provided, I think we need to implement mechanisms to filter out spent UTxOs. Do we want to do it now? I think this feature may be added in next version of Kupo, which may provide more elegant ways around it. Curious to hear your thoughts!

cffls: That is a great point, Henry. I didn't realize that kupo will return all transaction outputs, instead of unspent transaction outputs. Thanks for pointing that out! 
`utxos` from a chaincontext object should always return unspent transaction outputs, so I don't think we should return any spent outputs. Before the filter flag is introduced to kupo, a solution to fix this now is to make an additional call to ogmios utxo query by [tx output reference](https://ogmios.dev/mini-protocols/local-state-query/) to check if the utxo are still unspent. Once kupo provides a filter for unspent transaction output, we can directly use that flag.

henryyuanheng-wang:
Note: made modification to integration test - Kupo doesn't sync the genesis UTxO (needs further investigation and discussion with KtorZ) so right now passing the genesis utxo manually as input. The rest of the UTxO queries work as expected.

cffls:
This is awesome. I will merge the code right away. Thank you Henry!

PR #42, opened 2022-05-23
Store Transaction Output amounts as Values

astrojarred:
The purpose of this PR is to help make UTxO math a bit easier.
At the moment, one cannot add the amounts of `TransactionOutput`s together easily.

An example of why this might be useful is:

```python
# add outputs to a transaction with builder 

builder.outputs[0] += Value(coin=5_000_000)

# or even 
builder.outputs[0] += 5_000_000
```

The above code works only if the `outputs[0]` is already a `Value`.  If it's an `int` it won't work. The current workaround is to do something like:
```python
builder.outputs[0] = Value(coin=5_000_000) + builder.outputs[0]
```

so that the value is the class handling the addition, but it might be a bit easier from a UTxO math standpoint if amounts were always `Value`s (but could still be initialized with `int`s.

There might be a problem with one of the tests but I'm having trouble debugging it.

### DIFF
--- a/integration-test/test/test_all.py
+++ b/integration-test/test/test_all.py
@@ -278,7 +278,8 @@ class TestAll:
 
         non_nft_utxo = None
         for utxo in self.chain_context.utxos(str(taker_address)):
-            if isinstance(utxo.output.amount, int):
+            # multi_asset should be empty for collateral utxo
+            if not utxo.output.amount.multi_asset:
                 non_nft_utxo = utxo
                 break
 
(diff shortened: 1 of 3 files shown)


astrojarred:
OK yeah it looks like `test_not_enough_input_amount` is not able to deserialize `Value({'coin': 5000000, 'multi_asset': {}})` for some reason. @cffls I'm guessing it might be more obvious to you than to me as to why this is failing.

cffls:
It is not obvious to me either. 😆  Thanks for creating this PR! I will take it from here.

cffls:
The issue was in the unit test, not the new code. The failed test assumed the amount was always an integer, which is not the case after this commit.

astrojarred:
Makes total sense, thanks for updating me and patching it up!

PR #45, opened 2022-05-30
Add message signing and verification

astrojarred:
This PR adds functionality for both the signing and verification of messages as layed out in [CIP-0008](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0008/README.md). It also attempts to follow [CIP-0030](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0030) such that messages signed by both pycardano vkeys and dApp wallets could be verified in python. There is no consensus yet on the exact details of how this should work, and currently there are two approaches both of which are addressed here.

The aforementioned CIPs use COSE (CBOR Object Signing and Encryption), the algorithms for which are already written in a well-maintained python repository called [pycose](https://github.com/TimothyClaeys/pycose) (`poetry add cose`).

Later goals include generating and verifying JWT-like tokens generated by [web3-cardano-token](https://github.com/pyropy/web3-cardano-token).

## The current approach
I just wanted to get some code out, and my guess is that the current approach will need to go through significant changes to increase ease of use.

I've created a class called `Message` which accepts either `message: str` or `signed_message: Union[str, dict]`. If you want to sign a message with pycardano initialize the class with `message`, and then use `Message.sign()` to sign it.

If you want to verify an already-signed message. initialize it with `signed_message` and then use `Message.verify()`.

The parameter `signed_message` accepts both str and dict because as of now, there are two main ways of signing messages employed by browser wallets:

1. You include the wallet's verification key in the header of the signature (Eternl and Flint). the final result is a hex string.
2. You include the key separately and return a dict of the form `{"signature": signature, "key": key}`. (Nami)

In order to specify which one you want while signing or verifying, use the parameter `cose_key_separate: bool` to decide whether or not to attach it to the cose message header (case 1) or keep it separate (case 2). See the added tests for examples. (Couldn't think of a better name for this parameter...)

I feel like this isn't the most elegant solution for implementing this, but it's the best I could think of at the moment. Let me know if you have suggestions or ideas!

### DIFF
--- a/pycardano/cip30.py
+++ b/pycardano/cip30.py
@@ -1,0 +1,146 @@
+from typing import Optional, Union
+
+from cose.algorithms import EdDSA
+from cose.headers import KID, Algorithm
+from cose.keys import CoseKey
+from cose.keys.curves import Ed25519
+from cose.keys.keyops import SignOp, VerifyOp
+from cose.keys.keyparam import KpAlg, KpKeyOps, KpKty, OKPKpCurve, OKPKpD, OKPKpX
+from cose.keys.keytype import KtyOKP
+from cose.messages import CoseMessage, Sign1Message
+
+from pycardano.address import Address
+from pycardano.key import PaymentVerificationKey, SigningKey, VerificationKey
+from pycardano.network import Network
+
+
+def sign(
+    message: str,
+    signing_key: SigningKey,
+    attach_cose_key: bool = False,
+    network: Network = Network.MAINNET,
+):
+
+    # derive the verification key
+    verification_key = VerificationKey.from_signing_key(signing_key)
+
+    # create the message object, attach verification key to the header
+    msg = Sign1Message(
+        phdr={
+            Algorithm: EdDSA,
+            "address": Address(verification_key.hash(), network=network).to_primitive(),
+        },
+        payload=message.encode("utf-8"),
+    )
+
+    msg.uhdr = {"hashed": False}
+
+    if not attach_cose_key:
+        msg.phdr[KID] = verification_key.to_primitive()
+
+    # build the CoseSign1 signing key from a dictionary
+    cose_key = {
+        KpKty: KtyOKP,
+        OKPKpCurve: Ed25519,
+        KpKeyOps: [SignOp, VerifyOp],
+        OKPKpD: signing_key.payload,  # private key
+        OKPKpX: verification_key.payload,  # public key
+    }
+
+    cose_key = CoseKey.from_dict(cose_key)
+
+    msg.key = cose_key  # attach the key to the message
+
+    encoded = msg.encode()
+
+    # turn the enocded message into a hex string and remove the first byte
+    # which is always "d2"
+    signed_message = encoded.hex()[2:]
+
+    if attach_cose_key:
+        key_to_return = {
+            KpKty: KtyOKP,
+            KpAlg: EdDSA,
+            OKPKpCurve: Ed25519,
+            OKPKpX: verification_key.payload,  # public key
+        }
+
+        signed_message = {
+            "signature": signed_message,
+            "key": CoseKey.from_dict(key_to_return).encode().hex(),
+        }
+
+    return signed_message
+
+
+def verify(signed_message: Union[str, dict], attach_cose_key: Optional[bool] = None):
+    """
+    Verify the signature of a COSESign1 message and decode.
+    Supports messages signed by browser wallets or `Message.sign()`.
+    """
+
+    if attach_cose_key is None:
+        # try to determine automatically if the verification key is included in the header
+        if isinstance(signed_message, dict):
+            attach_cose_key = True
+        else:
+            attach_cose_key = False
+
+    if attach_cose_key:
+        # The cose key is attached as a dict object which contains the verification key
+        # the headers of the signature are emtpy
+        key = signed_message.get("key")
+        signed_message = signed_message.get("signature")
+
+    else:
+        key = ""  # key will be extracted later from the payload headers
+
+    # Add back the "D2" header byte and decode
+    decoded_message = CoseMessage.decode(bytes.fromhex("d2" + signed_message))
+
+    # generate/extract the cose key
+    if not attach_cose_key:
+
+        # get the verification key from the headers
+        verification_key = decoded_message.phdr[KID]
+
+        # generate the COSE key
+        cose_key = {
+            KpKty: KtyOKP,
+            OKPKpCurve: Ed25519,
+            KpKeyOps: [SignOp, VerifyOp],
+            OKPKpX: verification_key,  # public key
+        }
+
+        cose_key = CoseKey.from_dict(cose_key)
+
+    else:
+        # i,e key is sent separately
+        cose_key = CoseKey.decode(bytes.fromhex(key))
+        verification_key = cose_key[OKPKpX]
+
+    # attach the key to the decoded message
+    decoded_message.key = cose_key
+
+    signature_verified = decoded_message.verify_signature()
+
+    message = decoded_message.payload.decode("utf-8")
+
+    print(decoded_message.phdr["address"])
+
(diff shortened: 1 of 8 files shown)


pycardano/key.py
```diff
@@ -287,3 +298,147 @@ def generate(cls) -> StakeKeyPair:
...
+        self.verified = None
+
+        self.signature_verified = None
+        self.signing_address = None
+        self.addresses_match = None
+
+    def sign(
+        self,
+        signing_key: SigningKey,
+        verification_key: VerificationKey,
```
cffls: `verification_key` could be derive from signing key by `VerificationKey.from_signing_key(signing_key)`, so we don't need to ask users to provide it.

pycardano/key.py
```diff
@@ -287,3 +298,147 @@ def generate(cls) -> StakeKeyPair:
...
+            signed_message = {
+                "signature": signed_message,
+                "key": binascii.hexlify(CoseKey.from_dict(key_to_return).encode()),
+            }
+
+        self.signed_message = signed_message
+
+        return signed_message
+
+    def verify(self, cose_key_separate: Optional[bool] = None):
```
cffls: To make `verify` a standalone function, we will need to pass `signed_message` and (`verification_key` or `verification_key_hash`) to it.

test/pycardano/test_key.py
```diff
@@ -97,3 +98,58 @@ def test_key_save():
         SK.save(f.name)
         sk = PaymentSigningKey.load(f.name)
         assert SK == sk
+
```
cffls: Thanks for adding unit tests, this is awesome! A interesting test case we can add here is to sign a message with a signing key and verify the signed message with its public key :D

pycardano/key.py
```diff
@@ -287,3 +298,147 @@ def generate(cls) -> StakeKeyPair:
     @classmethod
     def from_signing_key(cls, signing_key: StakeSigningKey) -> StakeKeyPair:
         return cls(signing_key, StakeVerificationKey.from_signing_key(signing_key))
+
+
+class Message:
```
cffls: I am not sure if we need a class to hold `sign` and `verify`, because they are both stateless (same inputs always generate same outputs). It might be a good idea to make them as standalone functions for simplicity. Also, I think we should move these functions to a new module (my apologies), e.g. "cip/cip30", because they are implementations of a specific CIP. This gives us flexibility of adding new message signing implementations in the future in case a new CIP is proposed.

pycardano/key.py
```diff
@@ -287,3 +298,147 @@ def generate(cls) -> StakeKeyPair:
...
+        cose_key = CoseKey.from_dict(cose_key)
+        msg.key = cose_key  # attach the key to the message
+
+        msg.uhdr = {"hashed": False}
+
+        encoded = msg.encode()
+
+        # turn the enocded message into a hex string and remove the first byte
+        # which is always "d2"
+        signed_message = binascii.hexlify(encoded).decode("utf-8")[2:]
```
cffls: This can probably be simplified with `.hex()` from bytes type.

```python
binascii.hexlify(b"123").decode("utf-8")
'313233'
b"123".hex()
'313233'
```

cffls (commented):
Thank you @astrojarred for contributing this great feature! The code looks good in general, and great job in adding tests! 👍   I just had some comments regarding some interface design decisions, but they are pretty easy to change.

astrojarred:
Awesome, thanks so much for the detailed comments. Will take a look soon.

pycardano/key.py
astrojarred: I don't believe we need to pass in the vkey as it will either be included in the headers of the signature object (eternal case) or passed along in a dict with the signature (nami). We just need to check that the vkey used to sign matches the one that is included in the headers.

cffls: That's even better, thanks!

astrojarred:
@cffls Looks like somehow I messed up Codecov... not sure what happened

cffls:
> 

No worries. It was an intermittent error from codecov's side. I restarted the jobs and they are passing now.

pycardano/cip30.py
```diff
@@ -0,0 +1,146 @@
+from typing import Optional, Union
```
cffls: Thank you for moving this to a standalone module! One last request, could you create a dedicated folder `cip`, rename this file to `cip8.py`, and move it to the folder? Message signing is the main topic of CIP8, so it makes more sense to name this file `cip8.py`. Keeping all CIPs in one folder will make the project structure cleaner.

pycardano/cip30.py
```diff
@@ -0,0 +1,146 @@
...
+from pycardano.key import PaymentVerificationKey, SigningKey, VerificationKey
+from pycardano.network import Network
+
+
+def sign(
+    message: str,
+    signing_key: SigningKey,
+    attach_cose_key: bool = False,
+    network: Network = Network.MAINNET,
+):
```
cffls: Could you add return type hint as well as a string doc for this function?

pycardano/cip30.py
```diff
@@ -0,0 +1,146 @@
...
+
+        signed_message = {
+            "signature": signed_message,
+            "key": CoseKey.from_dict(key_to_return).encode().hex(),
+        }
+
+    return signed_message
+
+
+def verify(signed_message: Union[str, dict], attach_cose_key: Optional[bool] = None):
```
cffls: Please add a return type hint for this function.

astrojarred:
Phew! Let me know if you have any more comments or suggestions here!

Now it's simply just `cip30.sign()` and `cip30.verify()`

pycardano/cip30.py
```diff
@@ -0,0 +1,146 @@
...
+        verification_key = cose_key[OKPKpX]
+
+    # attach the key to the decoded message
+    decoded_message.key = cose_key
+
+    signature_verified = decoded_message.verify_signature()
+
+    message = decoded_message.payload.decode("utf-8")
+
+    print(decoded_message.phdr["address"])
```
cffls: Let's use logger from module `logging` instead. Something like this:

```python
from pycardano.logging import logger

logger.debug(decoded_message.phdr["address"])
```

cffls:
Thanks @astrojarred for taking my feedbacks. This looks great! Before merging it, I will look into the CIP again and make sure we are not missing anything in the code.

astrojarred:
Excellent idea, there could definitely be some updates that I've missed.

pycardano/cip30.py
```diff
@@ -0,0 +1,146 @@
...
+
+    signing_address = Address.from_primitive(decoded_message.phdr["address"])
+
+    # check that the address attached matches the
+    # one of the verification keys used to sign the message
+    addresses_match = (
+        signing_address.payment_part
+        == PaymentVerificationKey.from_primitive(verification_key).hash()
```
cffls: Is there a case where we also need to compare staking part?

astrojarred:
Thanks for the great suggestions, will take care of them soon!

pycardano/cip30.py
astrojarred: I did not intend to have this here. I was print debugging and forgot to take this out 😣

Will use the logging module for all future prints though!

pycardano/cip30.py
astrojarred: I don't think so-- at least not yet. dapp wallets should use the payment keys to sign messages as per CIP-0030. However, the CIP does mention:

> The staking key will be used for reward addresses. 

but that's the only time anything of this sort is addressed at the moment. I guess at some point there might be the ability to sign messages from a reward address using your staking keys, but it's not implemented yet. I say we wait on this.

cffls: Okay, thanks for the explanation. We can add staking part if needed in the future.

astrojarred: Sure, let's keep an eye on it!

cffls:
Thank you @astrojarred for the final version, which looks great to me! I will merge it right away. Sorry about the delay.

PR #46, opened 2022-06-06
Add usage guides for Plutus

henryyuanheng-wang:
- Add usage guides for Plutus - explained concepts such as script address, datum, redeemer, script context, etc. Demonstrated the use through FortyTwo example. 
- Minor fix to the hyperlink in Transaction.rst
- removed the "rm -r docs/build" commands before running Sphinx in Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,7 @@ format: ## runs code style and formatter
 
 docs: ## build the documentation
 	poetry export --dev --without-hashes > docs/requirements.txt
+	rm -r -f docs/build
 	poetry run sphinx-build docs/source docs/build/html
 	$(BROWSER) docs/build/html/index.html
 
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,6 @@ format: ## runs code style and formatter
 
 docs: ## build the documentation
 	poetry export --dev --without-hashes > docs/requirements.txt
-	rm -r docs/build
 	poetry run sphinx-build docs/source docs/build/html
 	$(BROWSER) docs/build/html/index.html
 
--- a/docs/source/guides/plutus.rst
+++ b/docs/source/guides/plutus.rst
@@ -1,0 +1,124 @@
+======
+Plutus
+======
+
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -66,7 +66,6 @@ format: ## runs code style and formatter
 
 docs: ## build the documentation
 	poetry export --dev --without-hashes > docs/requirements.txt
-	rm -r docs/build
```
cffls, line 69: Why remove this? IIRC, this was to clean up docs folder before rebuilding it.

docs/source/guides/plutus.rst
```diff
@@ -0,0 +1,124 @@
...
+    ... class Test(PlutusData):
+    ...     CONSTR_ID = 1
+    ...     a: int
+    ...     b: bytes
+
+    >>> test = Test(123, b"321")
+    >>> test.to_cbor()
+    'd87a9f187b43333231ff'
+    >>> assert test == Test.from_cbor("d87a9f187b43333231ff")
+    True
```
cffls: I feel it might be a good idea to move this to a different section. Having it in the middle of forty-two example could be a distraction to readers. We can also add more datum serialization examples in the new section. There are already many examples in the unit test file [here](https://github.com/cffls/pycardano/blob/main/test/pycardano/test_plutus.py#L19-L53).

cffls (commented):
Thanks Henry, this looks awesome! One small suggestion, we can probably create a dedicated section for datum/redeemer serialization.

docs/source/guides/plutus.rst
henryyuanheng-wang: Yep that's a great suggestion! Made the modification to the file.

Makefile
henryyuanheng-wang: I think if building doc for the first time, the remove step will complain. We can add a -f flag to remove folder if exists like this `rm -r -f docs/build`.  What do you think?

cffls (approved):
LGTM. I will rebase your changes and resolve the conflicts before merging.

PR #58, opened 2022-07-02
Policy json serializer

astrojarred:
Here is the outline for a native script dict/json parser. I'm curious if we're happy with the logic. Specifically I'm not sure if there is a more elegant way to refer to deal with `TAGS`, `TYPES`, and `FIELDS` as I have.

Also, do we want to add `to_json()` and `from_json()` or let people do that on their own like in the last example? 

Next steps:
- Manually test with all types of policies
- Write tests

Example:
You can take any native script (say [directly copied from here](https://pool.pm/policy/862cd06c4504de6114a29e0b863751ee84ad455493d43aeeb727d896)) and read it into pycardano like so:

```python
policy = {
  "type": "all",
  "scripts": [
    {
      "type": "sig",
      "keyHash": "cd5c585cd696cce6aa3cefdcfca2dc2f3dad15191cb2c29be41593d1"
    },
    {
      "type": "before",
      "slot": 49399461
    }
  ]
}

script = NativeScript.from_dict(policy)
# ScriptAll(_TYPE=1, native_scripts=[ScriptPubkey(_TYPE=0, key_hash=VerificationKeyHash(hex='cd5c585cd696cce6aa3cefdcfca2dc2f3dad15191cb2c29be41593d1')), InvalidHereAfter(_TYPE=5, after=49399461)])
```

or save the same script to a json file and:

```python
import json

with open("./policy.script", "r") as f:
    script_dict = json.loads(f)

script = NativeScript.from_dict(policy)
```

Finally, you can export any native script to a dict / json file like so:

```python
# .
# .
# .
"""Create dummy policy"""
# A policy that requires a signature from the policy key we generated above
pub_key_policy = ScriptPubkey(policy_vkey.hash())

# A time policy that disallows token minting after 10000 seconds from last block
must_before_slot = InvalidHereAfter(chain_context.last_block_slot + 10000)

# Combine two policies using ScriptAll policy
policy = ScriptAll([pub_key_policy, must_before_slot])

# export policy to dict
policy_dict = policy.to_dict()

# export to json file
with open("./policy.script", "w+") as f:
    json.dump(policy_dict, f, indent=4)
```


pycardano/nativescript.py
```diff
@@ -52,6 +52,72 @@ def hash(self) -> ScriptHash:
...
+            native_script = [cls.from_dict(i, top_level=False) for i in script]
+
+        if not top_level:
+            return native_script
+        else:
+            return super(NativeScript, TYPES[script["type"]]).from_primitive(
+                native_script[1:]
+            )
+
+    def to_dict(self) -> dict:
```
cffls, line 96: Seems like this is a top down approach, where `NativeScript` will convert each field to a dict. A small disadvantage of it is that we need to keep some mapping like `TAGS` and `FIELDS` in the function. A bottom up approach might be simpler, where each subclass, e.g. `ScriptPubkey`, `ScriptAll`, will implement its `to_dict`.

pycardano/nativescript.py
```diff
@@ -52,6 +52,72 @@ def hash(self) -> ScriptHash:
...
+    ]:
+
+        TYPES = {
+            "sig": ScriptPubkey,
+            "all": ScriptAll,
+            "any": ScriptAny,
+            "atLeast": ScriptNofK,
+            "after": InvalidBefore,
+            "before": InvalidHereAfter,
+        }
```
cffls: Each key of this dictionary could be moved to its corresponding policy, so we don't need to define them in two places. Something like this:


```python
@dataclass
class ScriptPubkey(NativeScript):
    json_tag: ClassVar[str] = "sig"
    _TYPE: int = field(default=0, init=False)
    key_hash: VerificationKeyHash


# In `from_dict()`:
types = {p.json_tag: p for p in [ScriptPubkey, ScriptAll, ScriptAny, ScriptNofK, InvalidBefore, InvalidHereAfter]}
```

cffls (commented):
Thanks for the contribution as always! Looks good in general. Added two comments about the specific implementation.

astrojarred:
Hi Jerry, thanks for the insightful comments! I tried a similar implementation at first but the extra fields I added were getting thrown into the output of to_primitive. It looks like maybe using `ClassVar` was the secret sauce I was missing :)

astrojarred:
Marked as ready for review, hopefully the new solution is more elegant.

Changes:
- Added class variables `json_tag` and `json_field` to each NativeScript child class.
- Added one-liner docstrings
- For `from_dict`: created a lookup dictionary from the class variables as suggested
- For `to_dict`: directly references the json class variables instead of creating lookup dicts
- Added tests

cffls (approved):
Looks good to me. Thank you @astrojarred !

PR #80, opened 2022-08-25
added support for new testnets in blockfrost

mateusap1:


mateusap1:
Not sure if this is the best approach to deal with the new testnets (preprod and review). Another alternative would be to call them all Network.TESTNET (since they share some "testnet" simmilarities like the address format), but require (or allow) user to specify the blockfrost URL.

mateusap1:
Yeah, I see address derivation is failing with my addition.
`Address(payment_vkey.hash(), stake_vkey.hash(), network=Network.PREPROD)` is giving a different value from `Address(payment_vkey.hash(), stake_vkey.hash(), network=Network.TESTNET)` which is not good
Maybe mantaining the same networks, but allowing a custom blockfrost url is actually the best option

cffls (approved):
Nice, thanks for the change!

PR #85, opened 2022-09-18
Hierarchical Deterministic Wallet Implementation

henryyuanheng-wang:
HDWallet that supports generation of private keys, public keys and chain codes for Shelley Address. Changes include:

- poycardano/hdwallet.py: HDWallet class definition and functions, including ingestion of seed, mnemonic words and entropy. Can derive keys based on index or derivation path.

- test/pycardano/test_hdwallet: test and verify address generation from implementation. Test cases are copied from cardano-seriliazation-lib. 

- pyproject.toml/poetry.lock: add mnemonic and ecpy dependencies.


robinboening:
Thanks for working on this! I haven't tried your branch, but by looking through the code it looks nice and clean to me 👍

pycardano/hdwallet.py
```diff
@@ -0,0 +1,485 @@
...
+        A = cv25519.encode_point(P)
+
+        # set root keys and parent keys
+        self._root_xprivate_key = self._xprivate_key = (kL, kR)
+        self._root_public_key = self._public_key = A
+        self._root_chain_code = self._chain_code = c
+
+        return self
+
+    def from_mnemonic(
```
cffls: I guess this is better to be a classmethod?

pycardano/hdwallet.py
```diff
@@ -0,0 +1,485 @@
...
+        self._root_public_key: Optional[bytes] = None
+        self._root_chain_code: Optional[bytes] = None
+        self._xprivate_key: Optional[tuple] = None
+        self._public_key: Optional[bytes] = None
+        self._chain_code: Optional[bytes] = None
+        self._path: str = "m"
+        self._depth: int = 0
+        self._index: int = 0
+
+    def from_seed(self, seed: str) -> "HDWallet":
```
cffls: Same as `from_mnemonic`, this could be a classmethod.

pycardano/hdwallet.py
```diff
@@ -0,0 +1,485 @@
...
+        cv25519 = Curve.get_curve("Ed25519")
+        k_scalar = int.from_bytes(kL, "little")
+        P = k_scalar * cv25519.generator
+        A = cv25519.encode_point(P)
+
+        self._xprivate_key = (kL, kR)
+        self._public_key = A
+        self._chain_code = c
+
+        return self
```
cffls: It may be more intuitive to return a new instance of `HDWallet` rather than changing the current one and returning `self`, so people can keep keys derived from different paths at the same time. What do you think?

test/pycardano/test_hdwallet.py
```diff
@@ -0,0 +1,70 @@
+from pycardano.address import Address
+from pycardano.hdwallet import HDWallet
+from pycardano.key import PaymentVerificationKey
+from pycardano.network import Network
+
+
+# Tests copied from: https://github.com/Emurgo/cardano-serialization-lib/blob/master/rust/src/address.rs
+
+MNEMONIC_12 = "test walk nut penalty hip pave soap entry language right filter choice"
+MNEMONIC_15 = "art forum devote street sure rather head chuckle guard poverty release quote oak craft enemy"
```
cffls: Does this support 24 words mnemonic?

cffls (commented):
Thanks Henry for raising this PR, this is really amazing! Looks pretty good in general. I added a few minor comments. 

One thing to consider is to integrate this with the existing [extended public key and signing key](https://github.com/cffls/pycardano/blob/main/pycardano/key.py#L176-L208) in PyCardano. I am thinking about adding a method `to_extended_signing_key()` in `HDWallet`. Don't feel the pressure to add it in this PR. We can do it in future PRs.

cffls:
One thing I forgot to mention, sorry.. since HDWallet was initially proposed in bip32, it may be a good idea to move this into the bip32 module [here](https://github.com/cffls/pycardano/blob/main/pycardano/crypto/bip32.py).

pycardano/hdwallet.py
henryyuanheng-wang: Thanks Jerry for taking the time to review! Yeah I was actually thinking of turning those into classmethods. I'll make the modifications.

test/pycardano/test_hdwallet.py
henryyuanheng-wang: Yep it supports 24 words too and I tested it. I'll add in the test case.

pycardano/hdwallet.py
henryyuanheng-wang: This is a a good question. I was thinking of keeping `derive_from_path` and `derive_from_index` to have similar behavior - for user to call `derive_from_index`, they may call the function multiple times, so I was thinking of modifying the keys in place while keeping the extended keys intact. Definitely interested in chatting more on this and figure out the most intuitive implementation :)

henryyuanheng-wang:
The following changes were made:
- modified  `from_seed`, `from_mnemonic` and `from_entropy` to class method
- return child hdwallet instance during derivation instead of overwriting the original hdwallet instance
- added a function to Key so that ExtendedSigningKey can be created from an HDWallet instance. I added function as part of `ExtendedSigningKey` class instead of a `HDWallet` class since it resides in bip32 crypto base file. So users should be able to initiate a signing key if they have a valid hdwallet instance. Let me know if this implementation makes sense or if there is a better place to integrate the function with existing code base. Thanks!

cffls (approved):
LGTM. Thank you Henry for the amazing work!

PR #89, opened 2022-09-28
Add wallet classes

astrojarred:
This is still a WIP but I wanted to get the ball rolling on this! Lots of basic functionality is now here, along with the first few tests and examples. See `/examples/wallet.py` for some examples of the most basic functionality. 

Current capabilities:
- [x] Generate keys
- [x] Load keys
- [x] Query utxos
- [x] Send ada
- [x] Send specific UTxOs
- [x] Get senders of all UTxOs
- [x] Get metadata for all held tokens
- [x] Get utxo block times and sort utxos
- [x] Mint / Burn tokens
- [x] Automatically load in token polices where wallet is a signer
- [x] Automatically create BlockFrost Chain Context (mainnet, preprod, and preview)
- [x] Attach messages to transactions
- [x] Sign messages
- [x] Add custom metadata fields
- [x] Multi-output transactions
- [x] Register wallet
- [x] Stake wallet
- [x] Withdraw rewards
- [x] Generate fully manual transactions that can do any / all of the above
- [x] Integrate strict type checking
- [ ] Write tests
- [ ] Add more examples

Future additions (possibly later PRs):
- [ ] Draft and sign multi-sig transactions
- [ ] Interaction with native scripts
- [ ] Interaction with plutus scripts
- [ ] Load wallets with mnemonic phrase (integrating `HDWallet`)

Now currently figuring out how to write tests for the transaction methods, using `test_transaction.py` for inspiration.


pycardano/wallet.py
```diff
@@ -0,0 +1,1934 @@
...
+    TransactionOutput,
+    UTxO,
+    Value,
+    Withdrawals,
+)
+from pycardano.txbuilder import TransactionBuilder
+from pycardano.utils import min_lovelace
+
+# set logging level to info
+logger.setLevel(logging.INFO)
```
cffls: Log level is supposed to set by the client. The library will simply use the logger and let client decide what level to use.

pycardano/wallet.py
```diff
@@ -0,0 +1,1934 @@
...
+
+
+class Amount:
```
cffls, line 68: Can we make this a frozen data class or at least make field `lovelace` and `ada` immutable? This will prevent people from changing on field without change the other.

pycardano/wallet.py
```diff
@@ -0,0 +1,1934 @@
...
+                            TokenPolicy(name=policy_id[:8], policy=policy_id),
+                            amount=amount,
+                            name=asset,
+                        )
+                    )
+
+        self._token_dict = tokens
+        self._token_list = token_list
+
+    def query_utxos(self, context: Optional[ChainContext] = None):
```
cffls: At the first glance, I thought `query_utxos` returns a list of utxos of the wallet, but after reading the code, I realized that it just sync utxos from chain context and returns nothing. What do you think of renaming this to `sync`, which is simpler and more precise about the behavior of this function?

pycardano/wallet.py
```diff
@@ -0,0 +1,1938 @@
...
+
+        return self.transact(
+            inputs=inputs,
+            outputs=Output(self, amount),
+            stake_registration=register,
+            delegations=pool_hash,
+            **kwargs,
+        )
+
+    def withdraw(
```
cffls: `withdraw` feels a bit ambiguous to me, because it could mean withdraw not only rewards, but anything in the wallet. Can we name it `withdraw_rewards`?

pycardano/wallet.py
```diff
@@ -0,0 +1,1938 @@
...
+    For more info see: https://docs.blockfrost.io/#tag/Cardano-Accounts/paths/~1accounts~1{stake_address}/get
+
+    Args:
+        stake_address (Union[str, Address]): The stake address to get the stake info of.
+        context (ChainContext): The context to use for the query. For now must be BlockFrost.
+
+    Returns:
+        dict: Info regarding the given stake address.
+    """
+    if isinstance(context, BlockFrostChainContext):
```
cffls: There are a few places using this pattern, where we check if the context is BF, else return nothing. This pattern could be wrapped in a python decorator, which will simply the code, something like this:


```python
from functools import wraps


def block_frost_only(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        if kwargs.get("context") == "block_frost":
            func(*args, **kwargs)
        else:
            print(f"Function {func.__name__} is only available for block_frost context.")

    return wrapper

@block_frost_only
def get_stake_info(stake_address, context=context):
    """This is docstring for first function"""
    print(f"Getting stake info for {stake_address} from {context}.")
```

Usage:
```python
>>> stake_address = "my_address"
>>> get_stake_info(stake_address, context="block_frost")
Getting stake info for my_address from block_frost.
>>> get_stake_info(stake_address, context="Ogmios")
Function get_stake_info is only available for block_frost context.
```

pycardano/wallet.py
```diff
@@ -0,0 +1,1938 @@
...
+
+        try:
+            return context.api.accounts(str(stake_address)).to_dict()
+        except ApiError:
+            return {}
+
+    else:
+        logger.warn(
+            "Retrieving stake address information is only possible with Blockfrost Chain Context."
+        )
```
cffls: Can we raise an exception when the wrong context is passed? Raising an explicit exception is probably safer than returning a None value. What do you think?

cffls (commented):
Thank you @astrojarred for this amazing work! This will make many things much easier to use for many people. The PR looks pretty good already. I added a few comments about the implementation and API naming. Please feel free to disagree. Regarding the failed CI tests, you can use `make format` to reformat the code and test it with `make qa`. Again, thank you so much for your contribution! 🙏

astrojarred:
Hi Jerry,

Thanks for the helpful review. I concur with everything suggested and I especially like using the decorator for blockfrost-only contexts. I'll make the changes soon. 

And my apologies about the formatting, I often forget 😅

Thanks again!

pycardano/wallet.py
astrojarred: Agreed. There are a few places where similar logic also might apply:

`Wallet.stake_info` and other stake-related methods warn when stake address is not yet registered. Not sure if it makes more sense to leave it as a warning and return nothing or raise and error of some sort.

`Wallet.sync()` warns when an address has not transacted on the chain yet. I think leaving it as a warning should be fine.
``

pycardano/wallet.py
astrojarred: Nice. I will implement like so, such that context as both an arg and a kwarg are checked, as well as raising an error when a different context type is provided.

```python
def blockfrost_only(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        if isinstance(kwargs.get("context"), BlockFrostChainContext) or isinstance(args[1], BlockFrostChainContext):
            func(*args, **kwargs)
        else:
            raise TypeError(f"Function {func.__name__} is only available for context of type BlockFrostChainContext.")

    return wrapper
```

pycardano/wallet.py
cffls: Makes sense, for these two cases we can use warning. Thank you!

astrojarred (commented):
In reference to commit [c30fba1](https://github.com/cffls/pycardano/pull/89/commits/c30fba131e85a0eab96ffcadf7617142bbe5a6af):
@cffls I have the code generate wallet keys using the format:

```
name.skey
name.vkey
name.stake.skey
name.stake.vkey
```

This is to make it easy for a wallet to load unique staking keys. Let me know what you think of this format.

Thinking about this more I now realize I should also add kwargs so one can directly specify the path to each for the four keyfiles individually.

cffls:
> In reference to commit [c30fba1](https://github.com/cffls/pycardano/pull/89/commits/c30fba131e85a0eab96ffcadf7617142bbe5a6af): @cffls I have the code generate wallet keys using the format:
> 
> ```
> name.skey
> name.vkey
> name.stake.skey
> name.stake.vkey
> ```
> 
> This is to make it easy for a wallet to load unique staking keys. Let me know what you think of this format.
> 
> Thinking about this more I now realize I should also add kwargs so one can directly specify the path to each for the four keyfiles individually.

The naming looks good to me. Thank you!

astrojarred:
Great, thanks! I believe I've addressed all the comments for now. Next step is to write more tests. I'll have to look in detail at the ones you've already written for how to properly test functions that are meant to submit transactions to the chain.

cffls:
> Great, thanks! I believe I've addressed all the comments for now. Next step is to write more tests. I'll have to look in detail at the ones you've already written for how to properly test functions that are meant to submit transactions to the chain.

Nice! You can take a look at [integration tests](https://github.com/cffls/pycardano/tree/main/integration-test), where a local chain will be created through docker compose and you can submit a bunch of real transactions to the chain.

daehan-koreapool:
FYI, the main branch now enforces stronger type hint in the Github ci/cd pipeline.

I would advise rebasing this branch on top of the latest main branch to ensure the newly added `wallet.py` module passes all static analyses.

astrojarred:
Done! I will work on straightening out the type hints, thank you!

pycardano/wallet.py
```diff
@@ -0,0 +1,1948 @@
...
+    @property
+    def bytes_name(self):
+        return bytes(self.name.encode("utf-8"))
+
+    @property
+    def policy_id(self):
+        return self.policy.policy_id
+
+    @blockfrost_only
+    def get_on_chain_metadata(self, context: ChainContext) -> dict:
```
daehan-koreapool: Hi Jarred, 

I see that this method expects `BlockFrostChainContext`.
Is it possible to retrieve the same asset metadata information from Ogmios as well?
If so, I think it would be better to enforce this method from `ChainContext` base class.

Also, I see that most of the methods requiring `ChainContext` dependency is restricted by `@blockfrost_only`.
If so, we could explicitly type all context parameters as `BlockFrostChainContext` and support these new methods directly from `BlockFrostChainContext` class.
By doing so, you wouldn't need to utilize this decorator anymore.

robinboening:
Hey Jarred,

thanks for the big chunk of new features! Looks very promising to me!

However, I'd like to bring one thing up for a discussion...

You are introducing a new class `Wallet` as part of the public API. I like that because it is abstracting a lot of things which makes it easy to use.

Now here is what caught my eye. For some weeks now we also have a class `HDWallet` https://github.com/Python-Cardano/pycardano/blob/88894e2cc97a8b01fed9a9fc99f2ff2861beac4b/pycardano/crypto/bip32.py#L78. Without looking into the details I would not understand why do we have `Wallet` and `HDWallet` and they don't provide the same / a similar interface.

First of all I am wondering what your thoughts are regarding the two in regards to the user/developer experience.

Secondly, I am wondering if it makes sense to extend the high-level `Wallet` class to support `HDWallet` keys (out of scope for this PR).

Happy to hear what you're going to say :)

pycardano/wallet.py
astrojarred: Hi Daehan, 

Thanks for the input! As far as I know the methods marked with `@blockfrost_only` cannot be queried with Ogmios as they need information from db-sync. However, it is true that with the new typing system you've implemented the decorator is not needed anymore. It also means that I can simplify the type checking in many other places as well. So thank you for that!

astrojarred:
Hi @robinboening,

Thanks for pointing out the `HDWallet` class. I think it would be a really nice idea to integrate into `Wallet`. I think some might appreciate the option to create a wallet using a mnemonic or having full control over the generation of keys, so I don't think it's out of scope for this class. I've updated the todo list above to specify the integration of `HDWallet`. In the end it might come in another PR later but I think it would be really nice to include.

astrojarred:
Quick update-- I've finally managed to integrate mypy so that the checks do not fail anymore. In addition I've gotten coverage up to ~44%. Still some tests to write but I'm closing in on this.
🔜

KINGH242:
How can I help with this? I can't see the logs for the failed checks as it was too long ago. I am working on a project that would benefit from having the wallet stuff abstracted away into PyCardano.

cffls:
@KINGH242 that's great! This PR is staled and I completely forget about it.. I think a few type checks are failing. If those are fixed we should be able to merge it.

astrojarred:
Hey all, I dropped the ball a bit on this one!  

I think I got stuck with test writing-- I recall that I mainly had trouble with the mocking of the blockfrost API and tx submission as it's something I had no experience with.

Also happy to help out wherever possible and finally get this done.

KINGH242:
Hey @astrojarred, I added some test and fixed a few things to allow it to work with the latest changes. I submitted a PR here https://github.com/astrojarred/pycardano/pull/33

PR #101, opened 2022-10-17
Fixing inconsistency between generated entropy value type and the expected HDWallet.entropy value type

daehan-koreapool:
`mnemonic` Python package generates an entropy value in `bytearray` type.
However, `HDWallet.entropy` class is expecting a string value and it is used as a string throughout the entire bip32.py module codebase.

This merge request explicitly converts generated entropy value as a string value to be consistent with the original implementer's intention.

### DIFF
--- a/pycardano/crypto/bip32.py
+++ b/pycardano/crypto/bip32.py
@@ -109,6 +109,9 @@ class HDWallet:
 
         Args:
             seed: Master key of 96 bytes from seed hex string.
+            entropy: Entropy hex string, default to ``None``.
+            passphrase: Mnemonic passphrase or password, default to ``None``.
+            mnemonic: Mnemonic words, default to ``None``.
 
         Returns:
             HDWallet -- Hierarchical Deterministic Wallet instance.
@@ -168,7 +171,7 @@ class HDWallet:
         return cls.from_seed(
             seed=hexlify(seed).decode(),
             mnemonic=mnemonic,
-            entropy=entropy,
+            entropy=unhexlify(entropy).decode("utf-8"),
             passphrase=passphrase,
         )
 
--- a/pycardano/crypto/bip32.py
+++ b/pycardano/crypto/bip32.py
@@ -171,7 +171,7 @@ class HDWallet:
         return cls.from_seed(
             seed=hexlify(seed).decode(),
             mnemonic=mnemonic,
-            entropy=unhexlify(entropy).decode("utf-8"),
+            entropy=hexlify(entropy).decode("utf-8"),
             passphrase=passphrase,
         )
 
--- a/pycardano/crypto/bip32.py
+++ b/pycardano/crypto/bip32.py
@@ -569,15 +569,13 @@ class HDWallet:
             )
         try:
             mnemonic = unicodedata.normalize("NFKD", mnemonic)
-            if language is None:
-                for _language in SUPPORTED_MNEMONIC_LANGS:
-                    valid = False
-                    if Mnemonic(language=_language).check(mnemonic=mnemonic) is True:
-                        valid = True
-                        break
-                return valid
-            else:
+            if language:
                 return Mnemonic(language=language).check(mnemonic=mnemonic)
+
+            for _language in SUPPORTED_MNEMONIC_LANGS:
+                if Mnemonic(language=_language).check(mnemonic=mnemonic) is True:
+                    return True
+            return False
         except ValueError:
             logger.warning(
                 "The input mnemonic words are not valid. Words should be in string format seperated by space."
--- a/pycardano/crypto/bip32.py
+++ b/pycardano/crypto/bip32.py
@@ -16,6 +16,8 @@ from typing import Optional
 from mnemonic import Mnemonic
 from nacl import bindings
 
+from pycardano.logging import logger
+
 __all__ = ["BIP32ED25519PrivateKey", "BIP32ED25519PublicKey", "HDWallet"]
 
 
@@ -581,7 +583,7 @@ class HDWallet:
             else:
                 return Mnemonic(language=language).check(mnemonic=mnemonic)
         except ValueError:
-            print(
+            logger.warning(
                 "The input mnemonic words are not valid. Words should be in string format seperated by space."
             )
 
@@ -600,4 +602,5 @@ class HDWallet:
         try:
             return len(unhexlify(entropy)) in [16, 20, 24, 28, 32]
         except ValueError:
-            print("The input entropy is not valid.")
+            logger.warning("The input entropy is not valid.")
+            return False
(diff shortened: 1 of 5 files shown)


daehan-koreapool:
```
-> print(entropy)
(Pdb) entropy
bytearray(b'\xdf\x9e\xd2^\xd1F\xbfC3j]|\xf79Y\x94')
(Pdb) unhexlify(entropy)
*** binascii.Error: Non-hexadecimal digit found
```
Apparently `entropy = Mnemonic(language="english").to_entropy(words=mnemonic)` code doesn't seem to be hex decodable binary data.

Storing `HDWallet.entropy` in bytes might be the only solution here.

pycardano/crypto/bip32.py
```diff
@@ -152,28 +155,18 @@ def from_mnemonic(cls, mnemonic: str, passphrase: str = "") -> HDWallet:
...
             seed=hexlify(seed).decode(),
             mnemonic=mnemonic,
-            entropy=entropy,
+            entropy=hexlify(entropy).decode("utf-8"),
             passphrase=passphrase,
         )
 
     @classmethod
-    def from_entropy(cls, entropy: str, passphrase: str = None) -> HDWallet:
+    def from_entropy(cls, entropy: str, passphrase: str = "") -> HDWallet:
```
daehan-koreapool, line 183: Explicitly requiring a string value for passphrase parameter.

A default `None` value raises an exception when provided into hashlib.pdkdf2_hmac() function for generating a seed value.

pycardano/crypto/bip32.py
```diff
@@ -152,28 +155,18 @@ def from_mnemonic(cls, mnemonic: str, passphrase: str = "") -> HDWallet:
             raise ValueError("Invalid mnemonic words.")
 
         mnemonic = unicodedata.normalize("NFKD", mnemonic)
-        passphrase = str(passphrase) if passphrase else ""
```
daehan-koreapool, line 155: No need to convert it to a string when the method parameter already explicitly requires an optional string type.

daehan-koreapool:
Since crypto/bip32.py is being implemented and maintained by PyCardano project contributers, I think this module should be included in the coverage report in my opinion.

What do you think @cffls ?

cffls:
> Since crypto/bip32.py is being implemented and maintained by PyCardano project contributers, I think this module should be included in the coverage report in my opinion.
> 
> What do you think @cffls ?

Thank you for raising this PR! Yes, I think it makes sense to include this file to coverage. Please feel free to change `.coveragerc`.

daehan-koreapool:
# Major Updates
* Ensuring Mnemonic package generated entropy value is stored as a string representation
* Pull out seed generating method for reusability
* Setting `from_entropy()` passphrase parameter as a string value by default
* Add more testcases ported from Emurgo serialization library to ensure both Mainnet and Testnet addresses are covered
* Add tests around entropy-based class methods to ensure refactored codes are functional

# Minor Updates
* Replace print statements with logging statements
* Update Makefile, gitignore, and coveragerc

cffls (approved):
LGTM, thanks for contributing to this project! @henryyuanheng-wang Henry, could you please take a look at this?

daehan-koreapool:
Hi @henryyuanheng-wang ,

I have a question regarding `test_payment_address_12_reward2()` test case.
For this test, I see that you are explicitly specifying `private=False` for the last two derivation path indexes.
From my experiment, setting this to `private=True` results in the same public key generation.
Would you be able to clarify on why this results in the same public key?

I've also just pushed another commit which can be rolled back if you wish to.
This commit is to enhance HDWallet creation UX to support simpler chaining of HDWallet derivation steps which is similar to other HDWallet library API design.

henryyuanheng-wang:
> Hi @henryyuanheng-wang ,
> 
> I have a question regarding `test_payment_address_12_reward2()` test case. For this test, I see that you are explicitly specifying `private=False` for the last two derivation path indexes. From my experiment, setting this to `private=True` results in the same public key generation. Would you be able to clarify on why this results in the same public key?
> 
> I've also just pushed another commit which can be rolled back if you wish to. This commit is to enhance HDWallet creation UX to support simpler chaining of HDWallet derivation steps which is similar to other HDWallet library API design.

Hi @daehan-koreapool , thanks for catching the inconsistencies and for making some great improvements! Everything looks pretty good. I like how you simplified the derivation step so a new HDWallet instance is created without needing to pass a parent HDWallet (was a bit redundant in hindsight). One question regarding the function name change from `derive_from_index` to `derive` - can you point me to the other popular implementation you were referring to? I was referencing some of the naming conventions from `python-hdwallet` [repo](https://github.com/meherett/python-hdwallet) but would love to see some other implementations. 

To answer your question regarding public keys derived either `private` or `public` is the same - I think this is expected behavior and beauty of HDWallet is that with public key from account you can derive the same public address as if you had the private key (you can refer to the diagram [here](https://input-output-hk.github.io/cardano-wallet/concepts/address-derivation) under HD Sequential wallets section and a HDWallet reading I found really useful [here](https://learnmeabitcoin.com/technical/hd-wallets)).  Let me know if it makes sense!

daehan-koreapool:
Hi @henryyuanheng-wang,

> Hi @daehan-koreapool , thanks for catching the inconsistencies and for making some great improvements! Everything looks pretty good. I like how you simplified the derivation step so a new HDWallet instance is created without needing to pass a parent HDWallet (was a bit redundant in hindsight). One question regarding the function name change from `derive_from_index` to `derive` - can you point me to the other popular implementation you were referring to? I was referencing some of the naming conventions from `python-hdwallet` [repo](https://github.com/meherett/python-hdwallet) but would love to see some other implementations.

It's pretty cool to know that derived public keys are identical for both private and public derivation methods.
Thanks for the clarification.

> To answer your question regarding public keys derived either `private` or `public` is the same - I think this is expected behavior and beauty of HDWallet is that with public key from account you can derive the same public address as if you had the private key (you can refer to the diagram [here](https://input-output-hk.github.io/cardano-wallet/concepts/address-derivation) under HD Sequential wallets section and a HDWallet reading I found really useful [here](https://learnmeabitcoin.com/technical/hd-wallets)). Let me know if it makes sense!

Regarding the `derive()` naming update, I referred to [`cardano-serialization-lib`](https://github.com/Emurgo/cardano-serialization-lib/blob/master/rust/src/address.rs#L1019) rust implementation.
This API design seem very simple and intuitive to use, so here's where I got the idea of returning a new HDWallet each time a derive() method is called for easier chaining operations.

[Lucid js library](https://doc.deno.land/https://deno.land/x/lucid@0.6.7/mod.ts/~/Core.Bip32PrivateKey
) also follows similar `derive(index)` pattern but this could be because Lucid uses `cardano-serialization-lib` wasm module under the hood.

daehan-koreapool:
I think the coverage test is failing since we are now including bip32.py module in the coverage report.
I'll add more testcases try to increase the test coverage.

daehan-koreapool:
Added a bit more refactoring commits as well.
* Pulled out frequently used list as a constant set. This could be set as a tuple if we want immutability, but since we use "contains" check more often, thought a set data structure is more suitable
* Cleaned up nested statements

PR #103, opened 2022-10-19
Testing more types of HDWallet derived Cardano addresses

daehan-koreapool:
Emurgo has Pointer and Enterprise address types included in their HDWallet derivation unit tests. 
https://github.com/Emurgo/cardano-serialization-lib/blob/master/rust/src/address.rs

PyCardano should have these address types covered to build a stronger assurance for other developers to reliable utilize the HDWallet implementation.


daehan-koreapool:
# Updates
* Added more enterprise and pointer address tests matching with Emurgo's serialization library
* Relocated bip32 tests to an appropriate directory

PR #107, opened 2022-10-25
Improve bip32.py type hint

daehan-koreapool:
Since I believe PyCardano will be a foundational and critical component of the future Python Cardano dev ecosystem, I would like to start improving PyCardano code quality by improving type hint one bit at a time.

# Goal
* improve bip32.py module's type hint enforcement
* possibly include mypy in the ci/cd pipeline without failure 

```
$ poetry run mypy pycardano/crypto/bip32.py
pycardano/crypto/bip32.py:134: error: Incompatible types in assignment (expression has type "bytearray", variable has type "str")
pycardano/crypto/bip32.py:135: error: Argument 1 to "_tweak_bits" of "HDWallet" has incompatible type "str"; expected "bytearray"
pycardano/crypto/bip32.py:335: error: Value of type "Optional[bytes]" is not indexable
pycardano/crypto/bip32.py:336: error: Value of type "Optional[bytes]" is not indexable
pycardano/crypto/bip32.py:344: error: Incompatible types in assignment (expression has type "Tuple[Optional[bytes], Optional[bytes], str]", variable has type "Tuple[Union[bytes, Any], Union[bytes, Any], Optional[bytes], Optional[bytes], str]")
pycardano/crypto/bip32.py:354: error: Syntax error in type annotation
pycardano/crypto/bip32.py:354: note: Suggestion: Use Tuple[T1, ..., Tn] instead of (T1, ..., Tn)
pycardano/crypto/bip32.py:396: error: Incompatible return value type (got "None", expected "HDWallet")
pycardano/crypto/bip32.py:444: error: Syntax error in type annotation
pycardano/crypto/bip32.py:444: note: Suggestion: Use Tuple[T1, ..., Tn] instead of (T1, ..., Tn)
pycardano/crypto/bip32.py:458: error: Incompatible return value type (got "None", expected "HDWallet")
pycardano/crypto/bip32.py:553: error: Missing return statement
Found 10 errors in 1 file (checked 1 source file)
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,9 +28,9 @@ jobs:
       - name: Install dependencies
         run: |
           poetry install
-      - name: Lint with flake8
+      - name: Run static analyses
         run: |
-          poetry run flake8 pycardano
+          make qa
       - name: Run unit tests
         run: |
           poetry run pytest --doctest-modules --ignore=examples --cov=pycardano --cov-config=.coveragerc --cov-report=xml
--- a/Makefile
+++ b/Makefile
@@ -60,8 +60,9 @@ test: ## runs tests
 test-single: ## runs tests with "single" markers
 	poetry run pytest -s -vv -m single
 
-qa: ## runs static analysis with flake8
+qa: ## runs static analyses
 	poetry run flake8 pycardano
+	poetry run mypy --install-types --non-interactive pycardano/crypto/bip32.py
 
 format: ## runs code style and formatter
 	poetry run isort .
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ test-single: ## runs tests with "single" markers
 
 qa: ## runs static analyses
 	poetry run flake8 pycardano
-	poetry run mypy --install-types --non-interactive pycardano/crypto/bip32.py
+	poetry run mypy --install-types --non-interactive pycardano
 
 format: ## runs code style and formatter
 	poetry run isort .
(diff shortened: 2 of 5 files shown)


daehan-koreapool:
# Major Updates
* Made most of the HDWallet constructor parameters non-optional typed since they are all used in each derivation step
* Identified there are no such cases where `public_pnode` and `private_pnode` can ever be empty

# Minor Updates
* Avoid type converting already defined variable
* Fixed type hint annotation syntax

cffls:
This is great, thanks for creating the first mypy check! One suggestion, could you also change this CI action to use mypy: https://github.com/cffls/pycardano/blob/main/.github/workflows/main.yml#L33? I think `make qa` is sufficient.

daehan-koreapool:
> This is great, thanks for creating the first mypy check! One suggestion, could you also change this CI action to use mypy: https://github.com/cffls/pycardano/blob/main/.github/workflows/main.yml#L33? I think `make qa` is sufficient.

Hi Jerry,

I've added a list of python modules to be excluded in the mypy and generalized `make qa` to target the entire pycardano package rather than a single bip32.py module. 
By doing so, any new modules added to the pycardano repo will need to pass mypy check, and we will be able to tackle one module at a time to improve type hint.

It seems there was an error on Codecov service while running the ci/cd pipeline.
I'm more familiar to Gitlab pipeline UX. Do you know how to re-trigger the pipeline step on Github?

cffls:
> I've added a list of python modules to be excluded in the mypy and generalized make qa to target the entire pycardano package rather than a single bip32.py module.
> By doing so, any new modules added to the pycardano repo will need to pass mypy check, and we will be able to tackle one module at a time to improve type hint.

Awesome, thanks for creating this smart check!

cffls:
Oops, missed your last question, sorry. You can go click on the failed job, and in the upper-right corner of the workflow, use the  Re-run jobs drop-down menu, and select Re-run all jobs. See this doc: https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs

PR #114, opened 2022-10-29
Refactor ogmios.py module maintainability

daehan-koreapool:
# Background
In my opinion, `pycardano.backend.base.GenesisParameters` and `pycardano.backend.base.ProtocolParameters` should be frozen dataclasses with no optional attributes.
I tested refactoring these dataclasses to require all field values, but it touches many of ogmios.py module.

# Goal
* Refactor ogmios.py as much as possible before `GenesisParameters` and `ProtocolParameters` become frozen dataclasses with full data requirements

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,7 @@ test-single: ## runs tests with "single" markers
 qa: ## runs static analyses
 	poetry run flake8 pycardano
 	poetry run mypy --install-types --non-interactive pycardano
+	poetry run black --check .
 
 format: ## runs code style and formatter
 	poetry run isort .
--- a/pycardano/address.py
+++ b/pycardano/address.py
@@ -9,7 +9,7 @@ Specifications and references could be found in:
 from __future__ import annotations
 
 from enum import Enum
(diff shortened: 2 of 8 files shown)


daehan-koreapool:
# Major Updates
* localizing Ogmosis query methods for reusability and for better maintainability
* adding more explicit return types for each Ogmosis query methods
* reordering methods so that helper methods are right beneath the caller method for easier top-to-bottom reading experience
* pulling out ProtocolParameter and GenesisParameter fetching logics so that it would be easier to update these dataclasses as non-optional + frozen dataclass in the next merge request

# Minor Updates
* Add `black` formatter check in the static analyses step

Makefile
```diff
@@ -63,6 +63,7 @@ test-single: ## runs tests with "single" markers
 qa: ## runs static analyses
 	poetry run flake8 pycardano
 	poetry run mypy --install-types --non-interactive pycardano
+	poetry run black --check .
```
cffls, line 66: Nice!

cffls (approved):

PR #117, opened 2022-10-29
Update variable name so it matches downstream code

peterVG:
Documentation correction: Line 172 requires a `signed_tx` variable but it's only defined as `tx` on line 151 resulting in an error.


cffls:
Thanks for fixing this!

PR #119, opened 2022-10-30
provide examples for adding transaction properties

peterVG:
closes issue#118

### DIFF
--- a/docs/source/guides/transaction.rst
+++ b/docs/source/guides/transaction.rst
@@ -142,6 +142,13 @@ Specify output amount::
 
 Step 6
 
+Add additional Plutus transaction properties as needed:
+
+    >>> builder.ttl = 3600
+    >>> builder.reference_inputs = [{"option1": "my_input1", "option2" : "my_input2"}]
+
+Step 7
+
 Create a signed transaction using transaction builder. Unlike building a raw transaction, where we need to manually
 sign a transaction and build a transaction witness set, transaction builder can build and sign a transaction directly
 with its `build_and_sign` method. The code below tells the builder to build a transaction and sign the transaction
--- a/docs/source/guides/transaction.rst
+++ b/docs/source/guides/transaction.rst
@@ -142,7 +142,7 @@ Specify output amount::
 
 Step 6
 
-Add additional Plutus transaction properties as needed:
+Add additional transaction information as needed:
 
     >>> builder.ttl = 3600
     >>> builder.reference_inputs = [{"option1": "my_input1", "option2" : "my_input2"}]
--- a/docs/source/guides/transaction.rst
+++ b/docs/source/guides/transaction.rst
@@ -145,7 +145,7 @@ Step 6
 Add additional transaction information as needed:
 
     >>> builder.ttl = 3600
-    >>> builder.reference_inputs = [{"option1": "my_input1", "option2" : "my_input2"}]
+    >>> builder.reference_inputs.add(tx_in)
 
 Step 7
 


docs/source/guides/transaction.rst
```diff
@@ -142,6 +142,13 @@ Specify output amount::
 
 Step 6
 
+Add additional Plutus transaction properties as needed:
+
+    >>> builder.ttl = 3600
+    >>> builder.reference_inputs = [{"option1": "my_input1", "option2" : "my_input2"}]
```
cffls: Not sure if this example will work out of the box. The right side of assignment should be a set of `TransactionInput`, right?

docs/source/guides/transaction.rst
```diff
@@ -142,6 +142,13 @@ Specify output amount::
 
 Step 6
 
+Add additional Plutus transaction properties as needed:
```
cffls: These are not "Plutus" per se, but ledger features added to Vasil era. I would just call these as additional transaction information.

docs/source/guides/transaction.rst
peterVG: It's working for me. Assuming that the `builder` object gets instantiated in Step 3 above I can see the following key:values in the transaction after running `builder.build_and_sign`

```
...
    'reference_inputs': [{'option1': 'my_input1', 'option2': 'my_input2'}],
    'required_signers': None,
    'script_data_hash': None,
    'total_collateral': None,
    'ttl': 3600,
    'update': None,
    'validity_start': None,
    'withdraws': None,
 ...
```

docs/source/guides/transaction.rst
peterVG: +1

docs/source/guides/transaction.rst
cffls: This line will run without error/exception, but it will run into problem when you submit this transaction to the network, because it doesn't conform to the ledger rule. I recommend to strictly define your reference input as `TransactionInput` type. Otherwise people following this guide will run into issue when they submit the tx.

peterVG: Ah understood. Yes, should have submitted to network to complete testing. Can you please provide an example @cffls of how to define my reference input example as a TransactionInput type? I thought I knew how that would work but getting errors.

cffls: Some examples could be found in this guide: https://pycardano.readthedocs.io/en/latest/guides/instance_creation.html

peterVG: I appreciate your patience @cffls but those examples still don't make it obvious to this novice Python programmer. Presumably many other potential PyCardano users (with limited Cardano experience) will have similar questions so I'm happy to submit more PR docs to add more explanatory context if I can figure this out.

What I do understand from existing docs, starting with the Instance/Object Creation page you've linked to:

We need to create an instance of [TransactionId] before creating a [TransactionInput]:
```
tx_id_hex = "732bfd67e66be8e8288349fcaaa2294973ef6271cc189a239bb431275401b8e5"
tx_id = TransactionId(bytes.fromhex(tx_id_hex))
```

Presumably the td_id_hex shown here is for a UTXO that will serve as input to this transaction?

Now I have a `tx_id` that can be used in a TransactionInput, like the following line in the docs example:
```
tx_in = TransactionInput(tx_id, 0)`
```

Except I want to use it to set the reference_inputs property for this transaction using my TransactionBuilder object:
```
builder = TransactionBuilder(context)
```

We've established that this won't work:
~~builder.reference_inputs = [{"option1": "my_input1", "option2" : "my_input2"}]~~

Instead,  "the right side of assignment should be a set of TransactionInput"

However, as per API docs, [TransactionInput ](https://pycardano.readthedocs.io/en/latest/api/pycardano.transaction.html#pycardano.transaction.TransactionInput) only has two parameters: transaction_id and index.

So where does the reference_inputs data go when using the TransactionBuilder? The examples only show it being used like this, with the two transaction_id and index parameters:

```
tx_in = TransactionInput(tx_id, 0)`
```

Obviously this doesn't work:
```
builder.reference_inputs = TransactionInput(tx_id, 0, reference_inputs=["testing"])
```

cffls: The way you construct tx inputs are correct. You just need to add `tx_in` to the `reference_inputs` field like this:

```python
tx_id_hex = "732bfd67e66be8e8288349fcaaa2294973ef6271cc189a239bb431275401b8e5"
tx_id = TransactionId(bytes.fromhex(tx_id_hex))
tx_in = TransactionInput(tx_id, 0)
builder.reference_inputs.add(tx_in)
```

peterVG: okay, almost there :-) thank you again for walking me through this.

So now where does the actual reference_inputs datum go? i.e. it could be a singular string value: "testing123", an integer value: 1234 or in many cases a dictionary: [{"option1": "my_input1", "option2" : "my_input2"}]

peterVG: Thanks and, sorry, it took me a moment to realize that reference_inputs actually only contain UTXOs and the actual content/data/info that a script needs to read in for its logic is stored in the inline datum for the UTXO that is linked in the reference_inputs field. So the datum has already been provided in a previous UTXO that we are referencing here. We are not providing the actual datum here in the reference_inputs field like we might do for the TTL field. Doh!

PR #128, opened 2022-11-09
Attempt to improve liskov substitution principle error

daehan-koreapool:
# Background
* Subclasses of `CBORSerializable` tend to override primitive related methods with tighter and specific parameter and return value types.
* Mypy raises the same error for most of the `CBORSerializable` subclasses and `network.py` module is the smallest module demonstrating the problem.
* According to Liskov Principle, subclasses should have looser type requirements rather than having tighter type requirements:
    * https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
    * https://stackoverflow.com/questions/56860/what-is-an-example-of-the-liskov-substitution-principle

# Goal
* Start a discussion in an attempt to improve PyCardano code quality
* I would love to hear your opinion on this, @cffls.

```
$ make qa
poetry run flake8 pycardano
pycardano/network.py:21: error: Return type "int" of "to_primitive" incompatible with return type "Primitive" in supertype "CBORSerializable"
pycardano/network.py:25: error: Argument 1 of "from_primitive" is incompatible with supertype "CBORSerializable"; supertype defines the argument type as "Primitive"
pycardano/network.py:25: note: This violates the Liskov substitution principle
pycardano/network.py:25: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
Found 2 errors in 1 file (checked 10 source files)
make: *** [Makefile:65: qa] Error 1

```

### DIFF
--- a/pycardano/serialization.py
+++ b/pycardano/serialization.py
@@ -9,7 +9,7 @@ from dataclasses import Field, dataclass, fields
 from datetime import datetime
 from decimal import Decimal
 from inspect import isclass
-from typing import Any, Callable, ClassVar, List, Type, TypeVar, Union, get_type_hints
+from typing import Any, Callable, List, Type, TypeVar, Union, get_type_hints
 
 from cbor2 import CBOREncoder, CBORSimpleValue, CBORTag, dumps, loads, undefined
 from pprintpp import pformat
@@ -53,8 +53,31 @@ class RawCBOR:
     cbor: bytes
 
 
-Primitive = TypeVar(
-    "Primitive",
+Primitive = Union[
+    bytes,
+    bytearray,
+    str,
+    int,
+    float,
+    Decimal,
+    bool,
+    None,
+    tuple,
+    list,
+    IndefiniteList,
+    dict,
+    defaultdict,
+    OrderedDict,
+    undefined.__class__,
+    datetime,
+    re.Pattern,
+    CBORSimpleValue,
+    CBORTag,
+    set,
+    frozenset,
+]
+
+PRIMITIVE_TYPES = (
     bytes,
     bytearray,
     str,
@@ -381,10 +404,10 @@ def _restore_dataclass_field(
                     return t.from_primitive(v)
                 except DeserializeException:
                     pass
-            elif t in Primitive.__constraints__ and isinstance(v, t):
+            elif t in PRIMITIVE_TYPES and isinstance(v, t):
                 return v
         raise DeserializeException(
-            f"Cannot deserialize object: \n{v}\n in any valid type from {t_args}."
+            f"Cannot deserialize object: \n{str(v)}\n in any valid type from {t_args}."
         )
     return v
 
@@ -453,8 +476,6 @@ class ArrayCBORSerializable(CBORSerializable):
         Test2(c='c', test1=Test1(a='a', b=None))
     """
 
-    field_sorter: ClassVar[Callable[[List], List]] = lambda x: x
-
     def to_shallow_primitive(self) -> List[Primitive]:
         """
         Returns:
@@ -465,7 +486,7 @@ class ArrayCBORSerializable(CBORSerializable):
                 types.
         """
         primitives = []
-        for f in self.__class__.field_sorter(fields(self)):
+        for f in fields(self):
             val = getattr(self, f.name)
             if val is None and f.metadata.get("optional"):
                 continue
@@ -473,7 +494,7 @@ class ArrayCBORSerializable(CBORSerializable):
         return primitives
 
     @classmethod
-    def from_primitive(cls: Type[ArrayBase], values: List[Primitive]) -> ArrayBase:
+    def from_primitive(cls: Type[ArrayBase], values: Primitive) -> ArrayBase:
         """Restore a primitive value to its original class type.
 
         Args:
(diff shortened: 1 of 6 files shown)


cffls:
Thanks @daehan-koreapool  for initiating this convo! Not sure why it is complaining line 21. The return type (`int`) is more restrictive than the super class (`Primitive`), even the example in [the guide](https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides) demonstrated that more restrictive return type is allowed. 

Regarding line 25, it makes sense to change the type of `value` to `Primitive`, and add a check at line 26 to make sure the value is an integer, otherwise raise an exception. This will fail early when the wrong input is passed in, and prevent obscure bugs.

daehan-koreapool:
I tried making bunch of changes on `serialization.Primitive` type since this is where most of the type warnings start from, and I noticed few things.

# Returning `TypeVar` without any method parameters raises an error
```
pycardano/serialization.py:175: error: A function returning TypeVar should receive at least one argument containing the same Typevar
pycardano/serialization.py:175: note: Consider using the upper bound "Union[bytes, bytearray, str, int, float, Decimal, bool, None, Tuple[Any, ...], List[Any], IndefiniteList, Dict[Any, Any], defaultdict[Any, Any], OrderedDict[Any, Any], datetime, Pattern[Any], Any, Any, Set[Any], FrozenSet[Any]]" instead
```

For a simple method which only returns a generic `Primitive` type without taking any parameters, mypy cannot infer what value to return. 

I also tried defining a generic type with a looser upper bound, but this raises the above error message once agin since we are still using a TypeVar.
```python3
Primitive = TypeVar(
    "Primitive",
    bound=Union[
        bytes,
        bytearray,
        str,
        int,
        float,
        Decimal,
        bool,
        None,
        tuple,
        list,
        IndefiniteList,
        dict,
        defaultdict,
        OrderedDict,
        # type(undefined),
        datetime,
        re.Pattern,
        CBORSimpleValue,
        CBORTag,
        set,
        frozenset,
    ],
)
```

# `Primitive` TypeVar generic type could be a `Union` type.
With a `TypeVar` generic type, only a single primitive type can exist in a function scope because how TypeVar works in Python.
Realistically, we should allow multiple CBOR primitive types to exist within a function scope and a `Union` type represents this better than more strict `TypeVar` generic type.

Also, when we define more simple `Primitive = Union[...]`, mypy now understands a method can return either of the primitive types.
So, the following error message is satisfied if we use an `Union` type:
```
pycardano/network.py:21: error: Return type "int" of "to_primitive" incompatible with return type "Primitive" in supertype "CBORSerializable"
```

>Regarding line 25, it makes sense to change the type of value to Primitive, and add a check at line 26 to make sure the value is an integer, otherwise raise an exception. This will fail early when the wrong input is passed in, and prevent obscure bugs.

Lastly, I do agree and mypy seems to be happy with this approach.

I'll try making few changes and push code changes for you to review through @cffls

cffls:
Thanks @daehan-koreapool , changing Primitive to a union type makes sense to me.

pycardano/serialization.py
```diff
@@ -381,10 +404,10 @@ def _restore_dataclass_field(
                     return t.from_primitive(v)
                 except DeserializeException:
                     pass
-            elif t in Primitive.__constraints__ and isinstance(v, t):
+            elif t in PRIMITIVE_TYPES and isinstance(v, t):
                 return v
         raise DeserializeException(
-            f"Cannot deserialize object: \n{v}\n in any valid type from {t_args}."
+            f"Cannot deserialize object: \n{str(v)}\n in any valid type from {t_args}."
```
daehan-koreapool, line 410: Had to explicitly convert Primitive variable into string representation since mypy was complaining a possible byte value will be represented as b"bytes value" when formatted with f-string.

pycardano/serialization.py
```diff
@@ -465,15 +486,15 @@ def to_shallow_primitive(self) -> List[Primitive]:
                 types.
         """
         primitives = []
-        for f in self.__class__.field_sorter(fields(self)):
+        for f in fields(self):
```
daehan-koreapool, line 489: Removed it since this lambda function didn't seem to have any real usages.

pycardano/serialization.py
```diff
@@ -718,13 +739,17 @@ def from_primitive(cls: Type[DictBase], value: dict) -> DictBase:
...
+                f"A dictionary value is required for deserialization: {str(value)}"
+            )
+
         restored = cls()
         for k, v in value.items():
             k = (
                 cls.KEY_TYPE.from_primitive(k)
-                if isclass(cls.VALUE_TYPE)
-                and issubclass(cls.KEY_TYPE, CBORSerializable)
+                if isclass(cls.KEY_TYPE) and issubclass(cls.KEY_TYPE, CBORSerializable)
```
daehan-koreapool, line 752: From my understanding, this line should check `KEY_TYPE` and not `VALUE_TYPE`.
Please let me know if this is supposed to stay as `VALUE_TYPE`.

pycardano/serialization.py
```diff
@@ -736,13 +761,13 @@ def from_primitive(cls: Type[DictBase], value: dict) -> DictBase:
             restored[k] = v
         return restored
 
-    def copy(self) -> DictBase:
+    def copy(self) -> DictCBORSerializable:
```
daehan-koreapool, line 764: Mypy wanted a concrete return type.

pycardano/serialization.py
cffls: Makes sense. It was needed in the past, but it is not used anymore. Thanks for cleaning this up.

pycardano/serialization.py
cffls: Nice catch! It was probably a copy-paste error 😅

cffls (approved):
Great work as always. Thank you @daehan-koreapool !

PR #130, opened 2022-11-14
Improve address type hint

daehan-koreapool:
# Goal:
* Improve address.py module's type hint related code quality

```
$ make qa
poetry run flake8 pycardano
pycardano/address.py:154: error: On Python 3 formatting "b'abc'" with "{}" produces "b'abc'", not "abc"; use "{!r}" if this is desired behavior  [str-bytes-safe]
pycardano/address.py:163: error: Argument 1 of "from_primitive" is incompatible with supertype "CBORSerializable"; supertype defines the argument type as "Union[bytes, bytearray, str, int, float, Decimal, bool, None, Tuple[Any, ...], List[Any], IndefiniteList, Dict[Any, Any], defaultdict[Any, Any], OrderedDict[Any, Any], Any, datetime, Pattern[Any], Any, Any, Set[Any], FrozenSet[Any]]"  [override]
pycardano/address.py:163: note: This violates the Liskov substitution principle
pycardano/address.py:163: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
pycardano/address.py:342: error: Argument 1 of "from_primitive" is incompatible with supertype "CBORSerializable"; supertype defines the argument type as "Union[bytes, bytearray, str, int, float, Decimal, bool, None, Tuple[Any, ...], List[Any], IndefiniteList, Dict[Any, Any], defaultdict[Any, Any], OrderedDict[Any, Any], Any, datetime, Pattern[Any], Any, Any, Set[Any], FrozenSet[Any]]"  [override]
pycardano/address.py:342: note: This violates the Liskov substitution principle
pycardano/address.py:342: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
pycardano/address.py:393: error: On Python 3 formatting "b'abc'" with "{}" produces "b'abc'", not "abc"; use "{!r}" if this is desired behavior  [str-bytes-safe]
Found 4 errors in 1 file (checked 13 source files)

```

### DIFF
--- a/pycardano/address.py
+++ b/pycardano/address.py
@@ -19,7 +19,7 @@ from pycardano.exception import (
 )
 from pycardano.hash import VERIFICATION_KEY_HASH_SIZE, ScriptHash, VerificationKeyHash
 from pycardano.network import Network
-from pycardano.serialization import CBORSerializable
+from pycardano.serialization import CBORSerializable, Primitive
 
 __all__ = ["AddressType", "PointerAddress", "Address"]
 
@@ -160,7 +160,11 @@ class PointerAddress(CBORSerializable):
         return self.encode()
 
     @classmethod
-    def from_primitive(cls: Type[PointerAddress], value: bytes) -> PointerAddress:
+    def from_primitive(cls: Type[PointerAddress], value: Primitive) -> PointerAddress:
+        if not isinstance(value, bytes):
+            raise DeserializeException(
+                f"A bytes value is required for deserialization: {value}"
+            )
         return cls.decode(value)
 
     def __eq__(self, other):
--- a/pycardano/address.py
+++ b/pycardano/address.py
@@ -343,7 +343,18 @@ class Address(CBORSerializable):
         return bytes(self)
 
     @classmethod
-    def from_primitive(cls: Type[Address], value: Union[bytes, str]) -> Address:
+    def from_primitive(cls: Type[Address], value: Primitive) -> Address:
+        if not isinstance(
+            value,
+            (
+                bytes,
+                str,
+            ),
+        ):
+            raise DeserializeException(
+                f"A bytes or a string value is required for deserialization: {value}"
+            )
+
         if isinstance(value, str):
             value = bytes(decode(value))
         header = value[0]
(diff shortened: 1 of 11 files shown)


daehan-koreapool:
# Updates
* disabling `str-bytes-safe` mypy check since this only affects exception messages formatting bytes value in f-string in pycardano
* `from_primitive()` now takes a generalized Primitive value and narrows down valid types within the function scope

pycardano/address.py
```diff
@@ -339,7 +343,18 @@ def to_primitive(self) -> bytes:
...
+        if not isinstance(
+            value,
+            (
+                bytes,
+                str,
+            ),
+        ):
+            raise DeserializeException(
+                f"A bytes or a string value is required for deserialization: {value}"
+            )
```
cffls: The pattern of checking if the input type is in a set of types seems to be used quite often. Is it possible to create a decorator for it to make the code cleaner?

test/pycardano/test_address.py
```diff
@@ -15,3 +18,27 @@ def test_payment_addr():
...
+class AddressTest(TestCase):
+    def test_from_primitive_invalid_value(self):
+        with self.assertRaises(DeserializeException):
+            Address.from_primitive(1)
+
+        with self.assertRaises(DeserializeException):
+            Address.from_primitive([])
+
+        with self.assertRaises(DeserializeException):
+            Address.from_primitive({})
```
cffls, line 44: Nice!

pycardano/address.py
daehan-koreapool: Yeah, that sounds perfect.
I'll creator a decorator to be used across the whole code-base.

daehan-koreapool:
I tried introducing `limit_primitive_type()` decorator class and removed all `if not isinstance(value, type)` repeated codes.
However, mypy wasn't smart enough to infer Primitive value type has been narrowed down and started raising bunch of Liskov errors again.

I think the cleanest solution is to declare `from_primitive()` base method to accept Any types of primitive values and just explicitly state allowed value types in overriding methods.
By doing so, both users and mypy understands what the expected input type is more clearly. 

Also, I am keeping `limit_primitive_type()` decorator as internal only and not included in `__all__` for now.

cffls (approved):
`limit_primitive_type` looks very neat. Thanks for the change!

PR #138, opened 2022-11-30
Modified IndefiniteList as subclass of UserList.

bhatt-deep:
To facilitate different list functionalities like append, pop, extend etc. in IndefiniteList, have implemented the IndefiniteList as a subclass of UserList from collections.

### DIFF
--- a/pycardano/serialization.py
+++ b/pycardano/serialization.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import re
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict, defaultdict, UserList
 from copy import deepcopy
 from dataclasses import Field, dataclass, fields
 from datetime import datetime
@@ -36,15 +36,8 @@ __all__ = [
 ]
 
 
-class IndefiniteList:
-    def __init__(self, items):
-        self.items = items
-
-    def __eq__(self, other):
-        if isinstance(other, IndefiniteList):
-            return self.items == other.items
-        else:
-            return False
+class IndefiniteList(UserList):
+    pass
 
 
 @dataclass
@@ -146,7 +139,7 @@ def default_encoder(
         # handling here to explicitly write header (b'\x9f'), each body item, and footer (b'\xff') to
         # the output bytestring.
         encoder.write(b"\x9f")
-        for item in value.items:
+        for item in value.data:
             encoder.encode(item)
         encoder.write(b"\xff")
     elif isinstance(value, RawCBOR):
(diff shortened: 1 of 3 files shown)


cffls:
Thanks! It is a great idea to use UserList. The code looks good to me. There are some errors in static type check. Once those are fixed, this PR should be good to go.

pycardano/serialization.py
```diff
@@ -146,7 +139,7 @@ def default_encoder(
         # handling here to explicitly write header (b'\x9f'), each body item, and footer (b'\xff') to
         # the output bytestring.
         encoder.write(b"\x9f")
-        for item in value.items:
+        for item in value.data:
```
cffls: We can just do `for item in value:` I think.

pycardano/serialization.py
```diff
@@ -240,7 +233,7 @@ def _dfs(value):
             elif isinstance(value, list):
                 return [_helper(k) for k in value]
             elif isinstance(value, IndefiniteList):
-                return IndefiniteList([_helper(k) for k in value.items])
+                return IndefiniteList([_helper(k) for k in value.data])
```
cffls: Same as above, `k in value` is probably going to work.

pycardano/serialization.py
bhatt-deep: Sure, will update that.

bhatt-deep:
@cffls I have updated the `value.data` to `value`. Then I was looking at the static types, The error is due to None type in Initlist of init method in UserList and we provide the Primitive types in the method while we call IndefiniteList,  
But need some suggestions on how to deal with this. 
 attaching the relevant code parts here. 
 ```
 class UserList(_collections_abc.MutableSequence):
    A more or less complete user-defined wrapper around list objects.

    def __init__(self, initlist=None):
        self.data = []
        if initlist is not None:
            # XXX should this accept an arbitrary sequence?
            if type(initlist) == type(self.data):
                self.data[:] = initlist
            elif isinstance(initlist, UserList):
                self.data[:] = initlist.data[:]
            else:
                self.data = list(initlist)
```
 and https://github.com/Python-Cardano/pycardano/blob/main/pycardano/serialization.py#L422-L427

**Comment by cffls** [comment]
> @cffls I have updated the `value.data` to `value`. Then I was looking at the static types, The error is due to None type in Initlist of init method in UserList and we provide the Primitive types in the method while we call IndefiniteList, But need some suggestions on how to deal with this. attaching the relevant code parts here.
> 
> ```
> class UserList(_collections_abc.MutableSequence):
>    A more or less complete user-defined wrapper around list objects.
> 
>    def __init__(self, initlist=None):
>        self.data = []
>        if initlist is not None:
>            # XXX should this accept an arbitrary sequence?
>            if type(initlist) == type(self.data):
>                self.data[:] = initlist
>            elif isinstance(initlist, UserList):
>                self.data[:] = initlist.data[:]
>            else:
>                self.data = list(initlist)
> ```
> 
> and https://github.com/Python-Cardano/pycardano/blob/main/pycardano/serialization.py#L422-L427

I guess the problem is that UserList is not using typehint. Maybe try creating a init method for IndefiniteList and ignore the type check when passing the value.

```python
class IndefiniteList(UserList):
    
    def __init__(list: List[Primitive]): # type: ignore
        super().__init__(list)
```

PR #143, opened 2023-01-13
Default to preprod testnet with blockfrost

nielstron:
An alternative would be to represent all networks in the NETWORK enum


cffls (approved):

cffls:
Thanks for the fix!

PR #144, opened 2023-01-13
Switch the Plutus introduction to eopsin

nielstron:
This changes the plutus example to an eopsin contract, which might help understanding the meaning of the contract.

### DIFF
--- a/docs/source/guides/plutus.rst
+++ b/docs/source/guides/plutus.rst
@@ -70,14 +70,51 @@ Similarly, redeemer can be serialized like following::
 Example - FortyTwo
 ------------------
 
-We demonstrate how these concepts come into play using a simple example from PPP - FortyTwo. The original script in haskell can be found here `here <https://github.com/input-output-hk/plutus-pioneer-program/blob/28559d379df8b66c06d8fbd1e2a43f6a8351382a/code/week02/src/Week02/Typed.hs>`_. Using PyCardano, we will show one can send and lock funds at a script address, and how someone else with the correct redeemer value can unlock and receive the funds.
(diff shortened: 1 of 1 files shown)


docs/source/guides/plutus.rst
```diff
@@ -70,14 +70,51 @@ Similarly, redeemer can be serialized like following::
...
+        sig_present = False
+        for s in context.tx_info.signatories:
+            if datum.pubkeyhash == s:
+                sig_present = True
+        assert sig_present
+
+
+Step 2
+
+Install the python packages ``eopsin-lang`` and ``pyaiken``. We can then build the contract.
```
cffls, line 100: If `pyaiken` is required to compile the script, could it be part of the dependencies in `eopsin-lang`?

cffls (commented):
Thank you for turning this example into pure python, love it! Just have a minor question regarding pyaiken.

docs/source/guides/plutus.rst
nielstron: The problem is that pyaiken is not pure python. There are many wheels for different distributions but the installation of eopsin should not fail due to pyaiken missing - rather users should use one of the alternative approaches to compile the contract (using Aiken directly for example)

If you know of other ways to ensure that pyaiken is available on any system I'll gladly make it a dependency of eopsin. I can also run some additional tests and see if it is automatically build in case no wheel is available.

nielstron: OK turns out that the wheel is built automatically if not available (and might require the user to install Rust). I will make pyaiken a normal dependency of eopsin then, so that it does not need to be installed separately.

cffls: Awesome, thank you for making it simpler!

cffls (approved):

PR #170, opened 2023-03-04
Correctly parse List[X] annotated objects

nielstron:
Imperatorlang allows users to specify PlutusData like this

```
@dataclass()
class D(PlutusData):
  xs: List[A]  # where A is also PlutusData
```

This was previously not correctly parsed by Pycardano. I suggest the List notation is also added to the main Pycardano, as it otherwise lacks the ability to build complex objects.


cffls:
Thanks! The change makes sense. Could you add a test for it? LGTM otherwise.

nielstron:
Future todos (for another PR)
- implement the same for Dict -> #180 
- ~~fix the case where CBORTags are > 7 (this is covered by the plutus implementation but not widely documented)~~

nielstron:
@cffls Please review

nielstron:
Again, coverage submission failed :/

cffls (approved):

PR #176, opened 2023-03-14
Fix timezone info for Ogmios backend

juliusfrost:
In the latest build, Ogmios gives timezone information in the return string. For example, `2022-10-25T00:00:00Z`


cffls:
Thank you for the fix! Looks like the unit test failed. To make it work again with any version of ogmios, maybe we can try the new parsing template first, and if parsing has failed, try removing `%z` from the template and parsing again. What do you think?

juliusfrost:
@cffls I changed it to use `datetime` instead, to make it more cohesive and it should handle both cases now.

cffls (approved):
LGTM. Thank you @juliusfrost !

PR #178, opened 2023-03-14
Remove the need to specify the RedeemerTag

### LINKED ISSUE #177: Specification of the Redeemer Tag necessary?
The Redeemer is currently annotated with a tag (SPEND, MINT, etc)

I am wondering if this is necessary? Should this tag not be automatically derived? When building transactions with the CLI, you never specify the tag. I would assume that the tag can be derived automatically depending on the position of its occurrence.

If you agree, I will open a PR that removes the need to specify this tag, reducing the friction for the TxBuilder.

nielstron:
This reduces friction for writing applications.

Also the proposed way of specifying a redeemer tag = MINT for minting inputs is actually invalid. Mint scripts can only be invoked without datums, by specifying them as minting scripts.

Fixes #177


cffls:
Thank you @nielstron for finding and removing this unnecessary complexity! Your change makes sense and it looks good to me.

cffls (approved):

PR #179, opened 2023-03-14
Add functions to automatically add required signers and validity range

nielstron:
Validity range and required signers are required by plutus scripts.
Users don't expect having to specify them if they add the utxos from specific addresses. Also validity range could be assumed to be somewhere around tx creation.

This PR adds the (default enabled) option to the transaction builder to automatically set the validity range and the required signers of the transaction to a reasonable value, which reduces friction for getting started with simple contracts.

### DIFF
--- a/pycardano/txbuilder.py
+++ b/pycardano/txbuilder.py
@@ -898,8 +898,8 @@ class TransactionBuilder:
 
         if auto_required_signers:
             # collect all signatories from explicitly defined transaction inputs and collateral inputs
-            required_signers = set(i.output.address.payment_part for i in self.inputs + self.collaterals)
-            self.required_signers = required_signers
+            required_signers = set(i.output.address.payment_part for i in self.inputs + self.collaterals if not isinstance(i.output.address.payment_part, ScriptHash))
+            self.required_signers = list(required_signers)
 
         selected_utxos = []
         selected_amount = Value()
--- a/pycardano/txbuilder.py
+++ b/pycardano/txbuilder.py
@@ -880,8 +880,10 @@ class TransactionBuilder:
             merge_change (Optional[bool]): If the change address match one of the transaction output, the change amount
                 will be directly added to that transaction output, instead of being added as a separate output.
             collateral_change_address (Optional[Address]): Address to which collateral changes will be returned.
-            auto_validity (Optional[bool]): Automatically set the validity interval of the transaction to a reasonable value
-            auto_required_signers (Optional[bool]): Automatically add all pubkeyhashes of transaction inputs to required signatories
+            auto_validity (Optional[bool]): Automatically set the validity interval of the transaction to a
+                reasonable value
+            auto_required_signers (Optional[bool]): Automatically add all pubkeyhashes of transaction inputs
+                to required signatories
 
         Returns:
             TransactionBody: A transaction body.
@@ -902,6 +904,7 @@ class TransactionBuilder:
                 i.output.address.payment_part
                 for i in self.inputs + self.collaterals
                 if not isinstance(i.output.address.payment_part, ScriptHash)
+                and i.output.address.payment_part is not None
             )
             self.required_signers = list(required_signers)
 
@@ -1201,8 +1204,10 @@ class TransactionBuilder:
             merge_change (Optional[bool]): If the change address match one of the transaction output, the change amount
                 will be directly added to that transaction output, instead of being added as a separate output.
             collateral_change_address (Optional[Address]): Address to which collateral changes will be returned.
-            auto_validity (Optional[bool]): Automatically set the validity interval of the transaction to a reasonable value
-            auto_required_signers (Optional[bool]): Automatically add all pubkeyhashes of transaction inputs to required signatories
+            auto_validity (Optional[bool]): Automatically set the validity interval of the transaction to
+                a reasonable value
+            auto_required_signers (Optional[bool]): Automatically add all pubkeyhashes of transaction inputs
+                to required signatories
 
         Returns:
             Transaction: A signed transaction.
--- a/pycardano/txbuilder.py
+++ b/pycardano/txbuilder.py
@@ -916,14 +916,6 @@ class TransactionBuilder:
                 auto_ttl_offset = 10_000
             self.ttl = max(0, last_slot + auto_ttl_offset)
 
-        # Automatically set the required signers for smart transactions
-        if (
-            is_smart and auto_required_signers is not False
-        ) and self.required_signers is None:
-            # Collect all signatories from explicitly defined
-            # transaction inputs and collateral inputs, and input addresses
-            self.required_signers = list(self._input_vkey_hashes())
-
         selected_utxos = []
         selected_amount = Value()
         for i in self.inputs:
@@ -1054,6 +1046,14 @@ class TransactionBuilder:
         )
 
         self.inputs[:] = selected_utxos[:]
+
+        # Automatically set the required signers for smart transactions
+        if (
+            is_smart and auto_required_signers is not False
+        ) and self.required_signers is None:
+            # Collect all signatories from explicitly defined
+            # transaction inputs and collateral inputs, and input addresses
+            self.required_signers = list(self._input_vkey_hashes())
 
         self._set_redeemer_index()
 
--- a/pycardano/txbuilder.py
+++ b/pycardano/txbuilder.py
@@ -869,8 +869,9 @@ class TransactionBuilder:
         change_address: Optional[Address] = None,
         merge_change: Optional[bool] = False,
         collateral_change_address: Optional[Address] = None,
-        auto_validity: Optional[bool] = True,
-        auto_required_signers: Optional[bool] = True,
+        auto_validity_start_offset: Optional[int] = None,
+        auto_ttl_offset: Optional[int] = None,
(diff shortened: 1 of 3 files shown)


nielstron:
cc @juliusfrost

pycardano/txbuilder.py
```diff
@@ -869,6 +869,8 @@ def build(
         change_address: Optional[Address] = None,
         merge_change: Optional[bool] = False,
         collateral_change_address: Optional[Address] = None,
+        auto_validity: Optional[bool] = True,
+        auto_required_signers: Optional[bool] = True,
```
cffls: I can see this making txbuilder easier to use, and it makes sense to add these parameters, but not sure if it is a good idea to enable them by default. Mainly two concerns:
1. Turning on these options will increase tx fee. Although it is not a large amount, I think we should try to make tx builder build the minimal tx whenever possible.
2. I could be wrong, but I think required signers are mostly used for phase 1 script validation. For regular transactions, it is unnecessary to add this field, because the error would've been caught anyway if any required signature is missing.

pycardano/txbuilder.py
```diff
@@ -878,11 +880,34 @@ def build(
...
         """
         self._ensure_no_input_exclusion_conflict()
+
+        if auto_validity:
+            # Automatically set the validity range to a tight value around transaction creation
+            last_slot = self.context.last_block_slot
+            if self.validity_start is None:
+                self.validity_start = max(0, last_slot - 1000)
+            if self.ttl is None:
+                self.ttl = last_slot + 10000
```
cffls: -1000 and +1000 seem a bit arbitrary to me. Can you make them configurable through the `build` method as well?

test/pycardano/util.py
```diff
@@ -90,7 +90,7 @@ def epoch(self) -> int:
         return 300
 
     @property
-    def slot(self) -> int:
+    def last_block_slot(self) -> int:
```
cffls, line 93: This would be a breaking change. Any strong reason why we want to change it?

nielstron: There is no method called "slot" in the super-class, so this method is likely unused. Also no method last_block_slot existed so I assumed it was just incorrectly named.

pycardano/txbuilder.py
nielstron: That's a good idea!

pycardano/txbuilder.py
nielstron: I would argue for the exact opposite. We want to make development as frictionless as possible. 

1. If you have to enable these parameters manually, you can also set them manually dkrectly

2. I have seen and experienced confusion about missing signatories and ranges so often now. Leaving them out should be a feature for adepts that want to save a few Lovelace, but for new users, the defaults should make it easy to use.

A good middle ground might be to just enable them per default if scripts are involved in the transaction?

test/pycardano/util.py
cffls: You are right, my bad. The base class declares `last_block_slot` instead of `slot`. 😅

pycardano/txbuilder.py
cffls: > A good middle ground might be to just enable them per default if scripts are involved in the transaction?

+100 Let's do this way!

nielstron:
@cffls The only error was that codecov could not upload the results, could you re-run?

Other than that, I addressed your changes, please re-review :)

pycardano/txbuilder.py
```diff
@@ -878,11 +881,60 @@ def build(
...
+                i.output.address for i in self.inputs + self.collaterals
+            ] + [
+                Address.from_primitive(a) if isinstance(a, str) else a
+                for a in self.input_addresses
+            ]
+            required_signers = set(
+                a.payment_part
+                for a in input_addresses
+                if isinstance(a.payment_part, VerificationKeyHash)
+            )
```
cffls: You can probably use `_input_vkey_hashes` here. https://github.com/Python-Cardano/pycardano/blob/db29ee4dcdc81fd928543c19402e23b255dd6019/pycardano/txbuilder.py#L661-L666

cffls (commented):
Thank you! Looks good overall. Just one minor comment about the computation of required signers.

nielstron:
I don't really understand why the continuous integration failed there, do you see it @cffls ?

cffls:
Integration test is non-deterministic. It will fail like 5% of time. Looks like retrying works.

pycardano/txbuilder.py
```diff
@@ -878,11 +881,58 @@ def build(
...
+        # Automatically set the required signers for smart transactions
+        if (
+            is_smart and auto_required_signers is not False
+        ) and self.required_signers is None:
+            # Collect all signatories from explicitly defined
+            # transaction inputs and collateral inputs, and input addresses
+            input_addresses = [
+                Address.from_primitive(a) if isinstance(a, str) else a
+                for a in self.input_addresses
+            ]
```
cffls: `self.input_addresses` is not necessary equivalent to the final selected utxo inputs. See the doc [here](https://github.com/Python-Cardano/pycardano/blob/main/pycardano/txbuilder.py#L292-L296).

`self.inputs + self.collaterals` used in `_input_vkey_hashes` is sufficient to cover all required signatures. However, I think we should move this part to a place after the utxo selection is completed, e.g. after this line: https://github.com/Python-Cardano/pycardano/blob/main/pycardano/txbuilder.py#L1026

nielstron: I assumed that. I also assumed that all additional input UTxOs would only come from addresses specified as input_addresses. Is this assumption incorrect?

I want to cover all addresses that were specified by the user explicitly as inputs, i.e. the input addresses and the inputs and collaterals.

cffls: > all additional input UTxOs would only come from addresses specified as input_addresses

This is incorrect. The final selected UTxOs will come from `self.inputs`. `self.input_addresses` is a backup solution when the sum of `self.inputs` is less than the sum of outputs. The correct way to describe `self.input_addresses` is that if an address is in `self.input_addresses`, it will have a chance to be included in final inputs `self.inputs`, which is the strict list of inputs the builder will use to build the final transaction.

```
                   ┌──────────────┐
                   │              │
          ┌────────┤              │
          │        │   txbuilder  │
  build   │        │              │
  final tx│        │              │
          │        └──────────────┘
          │
          │
┌─────────▼──────────┐
│                    │
│    self.inputs     │
│                    │
│         ┌──────────┼────────────────┐
│         │          │                │
│         │          │                │
│         │          │                │
│         │          │                │
└─────────┼──────────┘                │
          │                           │
          │   self.input_addresses    │
          │                           │
          └───────────────────────────┘
```

nielstron: I think I still understood it right. In the current code, there is no way that, there could be a UTxO spent from an address A where the verification key hash of A is not kn required signers.

Anyways, if you prefer, I can move the invocation to after the inputs selection.

nielstron: I would however always use all addresses specified in input_addresses as required signers because the user explicitly specified these addresses to the txbuilder, and would hence assume the corresponding key hash to be among the required signers.

cffls: Even when the user explicitly specified these addresses, they won't necessarily get selected. For example, a user has two addresses under an HD wallet. Each address contains 10 ADA. Let's assume the user doesn't know how much ADA is owned by each address, so she just passes both addresses to `self.add_input_address`. Now, if the user wants to send out 5 ADA, only the first address will be used by the UTxO selector, and the second address won't get selected, because the first address is able to cover the full transaction. 
On the other hand, if the user really want to make sure all specified UTxOs are included as tx inputs, `self.add_input` should be used instead of `self.add_input_address`.

nielstron: I understand that not all addresses are necessarily selected. But that's not the point.

The point is that the user added the address to the txbuilder as input address, so they would reasonably assume that the address is added as a required signature (/visible as signature to the contract)

cffls: I can't fully agree with this. I guess this boils down to whether we want to build the minimum possible transaction. In an extreme case where an HD wallet contains thousands of addresses, I am not sure if the user is okay with signing and adding that amount of data to the transaction. The beauty of `add_input_address` is that users can pass as many addresses as possible without worrying about the size of transaction or understanding the balance of each address.

nielstron: I see... in that case I would just leave out input_addresses and only take into account inputs that were explicitly added.

The utxos selected from the input addresses are almost unpredictable to the user so it would be safer to just not add them to the required signers.

pycardano/txbuilder.py
```diff
@@ -878,11 +881,49 @@ def build(
...
+                auto_ttl_offset = 10_000
+            self.ttl = max(0, last_slot + auto_ttl_offset)
+
+        # Automatically set the required signers for smart transactions
+        if (
+            is_smart and auto_required_signers is not False
+        ) and self.required_signers is None:
+            # Collect all signatories from explicitly defined
+            # transaction inputs and collateral inputs, and input addresses
+            self.required_signers = list(self._input_vkey_hashes())
```
cffls: Thanks! I would still move this part to below, because at this point, some utxo may have not been selected. I will pull your change and work on it if it is too much ask.

cffls (approved):
LGTM. Thank you!

PR #180, opened 2023-03-14
Add support for complex dictionary types

nielstron:
Same as #170 but for Dict[X, Y]

Note that per default, the plutus dataclasses of pycardano are not frozen and can hence not be used as key values. Technically, this is allowed in plutus.

### DIFF
--- a/pycardano/plutus.py
+++ b/pycardano/plutus.py
@@ -589,6 +589,35 @@ class PlutusData(ArrayCBORSerializable):
                                 raise DeserializeException(
                                     f"Unexpected data structure: {f}."
                                 )
+                        elif (
+                            hasattr(f_info.type, "__origin__")
+                            and f_info.type.__origin__ is dict
+                        ):
+                            t_args = f_info.type.__args__
+                            if len(t_args) != 2:
+                                raise DeserializeException(
+                                    "Dict type with wrong number of arguments"
+                                )
+                            if "map" not in f:
+                                raise DeserializeException(
+                                    f'Expected type "map" in object but got "{f}"'
+                                )
+                            key_t = t_args[0]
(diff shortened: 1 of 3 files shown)


pycardano/serialization.py
```diff
@@ -37,6 +37,10 @@
 ]
 
 
+def identity(x):
```
cffls: For the utility function used internally, I'd give it a `_` prefix to make it a "private" function.

cffls:
Awesome! Wanted to add this since a long time ago and you finally did it, haha.

pycardano/plutus.py
```diff
@@ -589,6 +589,35 @@ def _dfs(obj):
...
+                                key_convert = key_t.from_dict
+                            else:
+                                key_convert = _dfs
+                            if inspect.isclass(val_t) and issubclass(val_t, PlutusData):
+                                val_convert = val_t.from_dict
+                            else:
+                                val_convert = _dfs
+                            converted_fields.append(
+                                {
+                                    key_convert(pair["k"]): val_convert(pair["v"])
```
cffls, line 617: For my knowledge, does this work if `key_t` is a PlutusData type? How would the dictionary hash any non-primitive data types?

cffls: Ah I see, we can make the child class of PlutusData a frozen class and set eq to true: https://stackoverflow.com/questions/52390576/how-can-i-make-a-python-dataclass-hashable

nielstron: Yes it only works for frozen dataclasses (but it works there)

cffls (approved):
LGTM, thank you! However, there are some conflicts need to be resolved.

PR #181, opened 2023-03-16
Reference UTxOs are UTxOs (not TransactionInputs)

nielstron:
Reference UTxOs have been so much requested by Smart Contract developers because they grant the Smart Contract access to the _content_ (i.e. corresponding output, datums, scripts) of the referenced input.

This implies that the whole UTxO is actually required to build the ScriptContext properly. This PR changes the structure of the TxBuilder class such that it stores the entire reference UTxO instead of just the TransactionInput, enabling generating the ScriptContext for redeemers from it (see https://github.com/OpShin/eopsin/pull/92)

### DIFF
--- a/pycardano/txbuilder.py
+++ b/pycardano/txbuilder.py
@@ -50,6 +50,7 @@ from pycardano.transaction import (
     MultiAsset,
     Transaction,
     TransactionBody,
+    TransactionInput,
     TransactionOutput,
     UTxO,
     Value,
@@ -105,7 +106,9 @@ class TransactionBuilder:
 
     withdrawals: Optional[Withdrawals] = field(default=None)
 
-    reference_inputs: List[UTxO] = field(init=False, default_factory=lambda: list())
+    reference_inputs: Set[Union[UTxO, TransactionInput]] = field(
+        init=False, default_factory=lambda: set()
+    )
 
     _inputs: List[UTxO] = field(init=False, default_factory=lambda: [])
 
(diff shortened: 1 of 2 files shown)


pycardano/txbuilder.py
```diff
@@ -106,9 +105,7 @@ class TransactionBuilder:
 
     withdrawals: Optional[Withdrawals] = field(default=None)
 
-    reference_inputs: Set[TransactionInput] = field(
-        init=False, default_factory=lambda: set()
-    )
+    reference_inputs: List[UTxO] = field(init=False, default_factory=lambda: list())
```
cffls: This will be a backward-imcompatible change. The reason of using set was to avoid input duplication. What do you think of changing the type to `Set[Union[TransactionInput, UTxO]]` or `Union[Set[TransactionInput], Set[UTxo]]` instead?

nielstron: I resolve the duplication when building the tx. I think UTxOs are not hashable either.

cffls: UTxOs are in fact hashable: https://github.com/Python-Cardano/pycardano/blob/a065bcea465aaeb9aa766334882eaeb5b3203b91/pycardano/transaction.py#L467-L470

It would be less amount of change to keep using Set as well.

nielstron: OK I will make it a Set of UTxOs.

If we allow Union[TransactionInput, UTxO] we need to filter out again later too. Also building the ScriptContext fails then. But it could be desirable if someone just wants to make sure another transaction is in the UTxO while this tx is being processed.

cffls (approved):
LGTM. Thank you for making this change!

PR #206, opened 2023-04-14
Simplifies the backend interface (use available classes!)

nielstron:
This is a minimal effort change to the backends to leverage the available classes in PyCardano for user-friendliness

In particular, if a class for something is available then the backend should support it


nielstron:
As you can see from the code changes, the number of occurrences where the code is called with cbor instead of a Transaction object is actually 0. Almost the same for addresses.

cffls (approved):
Thanks! LGTM. This is a breaking change. Will include it in v0.9.

PR #208, opened 2023-04-16
Validate field type on serializing

cffls:
Validate each field in CBORSerializable during serialization. This will avoid bugs caused by incorrect type of field. For instance, when a field should be Address type, but being a string type instead, it would've been serialized successfully but failed to be deserialized.

This validation also helped discovering two minor type mismatch in unit tests. 🥳

### DIFF
--- a/pycardano/serialization.py
+++ b/pycardano/serialization.py
@@ -11,7 +11,18 @@ from datetime import datetime
 from decimal import Decimal
 from functools import wraps
 from inspect import isclass
-from typing import Any, Callable, List, Optional, Type, TypeVar, Union, get_type_hints
+from typing import (
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    List,
+    Optional,
+    Type,
+    TypeVar,
+    Union,
+    get_type_hints,
+)
 
 from cbor2 import CBOREncoder, CBORSimpleValue, CBORTag, dumps, loads, undefined
 from pprintpp import pformat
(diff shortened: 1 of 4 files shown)


nielstron:
Very neat! I like how this also applies to everything that is serialized very elegantly.

nielstron:
I just tried this with the error I initially reported. Its a better error message for sure, but there is still some improvement possible :)

```python
    builder.required_signers = ["string!"]
```

results in

```
TypeError: Field 'required_signers' should be of type typing.Union[typing.List[pycardano.hash.VerificationKeyHash], 
NoneType], got <class 'list'> instead.
```

The error here is that I have a list of "str" and not list of the required type, which upsets the type checker - it fails to point out the precise violation though.

I think the only reasonable fix for this would be a recursive type printer. Will look into that.

nielstron:
Suggestion: Instead of printing the type of the violating object (which may hide recursive errors), just print the violating value? This also helps the developer debug, as they might have a better grip on where a value comes from than what type it has.

Changes the error message to
```
TypeError: Field 'required_signers' should be of type typing.Union[typing.List[pycardano.hash.VerificationKeyHash], 
NoneType], got ['string!'] instead.
```

pycardano/serialization.py
```diff
@@ -254,7 +265,45 @@ def validate(self):
...
+                        _check_recursive(item, arg) for item, arg in zip(value, args)
+                    )
+            return True  # We don't know how to check this type
+
+        for field_name, field_type in type_hints.items():
+            field_value = getattr(self, field_name)
+            if not _check_recursive(field_value, field_type):
+                raise TypeError(
+                    f"Field '{field_name}' should be of type {field_type}, "
+                    f"got {type(field_value)} instead."
```
nielstron:
```suggestion
                    f"got {repr(field_value)} instead."
```

cffls: Makes sense. Thanks for the suggestion!

PR #211, opened 2023-04-16
Per default refetch the chain tip at most every 20 seconds

nielstron:
At the current state, every access to protocol params causes at least one chain tip fetch. Due to the potential delay of this request, often another protocol params fetch is triggered as well.

This significantly (and mostly unnecessarily) slows down transaction building. This fix caches the chain tip for a default of 20 seconds (the average slot time), making sure that subsequent calls to _is_chain_tip_updated return without any delay, drastically speeding up tx building.

### DIFF
--- a/pycardano/backend/ogmios.py
+++ b/pycardano/backend/ogmios.py
@@ -56,17 +56,19 @@ class OgmiosChainContext(ChainContext):
         network: Network,
         compact_result=True,
         kupo_url=None,
-        refetch_chain_tip_interval=20,
+        refetch_chain_tip_interval: Optional[int]=None,
     ):
         self._ws_url = ws_url
         self._network = network
         self._service_name = "ogmios.v1:compact" if compact_result else "ogmios"
         self._kupo_url = kupo_url
         self._last_known_block_slot = 0
-        self._refetch_chain_tip_interval = refetch_chain_tip_interval
+        self._refetch_chain_tip_interval = refetch_chain_tip_interval if refetch_chain_tip_interval is not None else 1000
         self._last_chain_tip_fetch = 0
         self._genesis_param = None
         self._protocol_param = None
+        if refetch_chain_tip_interval is None:
+            self._refetch_chain_tip_interval = self.genesis_param.slot_length / self.genesis_param.active_slots_coefficient
 
     def _request(self, method: OgmiosQueryType, args: JsonDict) -> Any:
         ws = websocket.WebSocket()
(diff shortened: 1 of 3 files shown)


nielstron:
Using the [opshin-starter-kit](https://github.com/OpShin/opshin-starter-kit/blob/main/src/off_chain/make_vest.py) as an example

Before
```
(opshin-starter-kit-py3.8) ➜  opshin-starter-kit git:(main) ✗ time python3 src/off_chain/make_vest.py owner beneficiary
transaction id: 4a26f76b7354d0d0f1c232dc187822701a1fe3faa565ffdc94a08ff9282eb51f
Cardanoscan: https://preview.cardanoscan.io/transaction/4a26f76b7354d0d0f1c232dc187822701a1fe3faa565ffdc94a08ff9282eb51f
python3 src/off_chain/make_vest.py owner beneficiary  0.60s user 0.08s system 1% cpu 43.842 total
```

After
```
(opshin-starter-kit-py3.8) ➜  opshin-starter-kit git:(main) ✗ time python3 src/off_chain/make_vest.py owner beneficiary
transaction id: 7042a915046c87c8cd1d7b953f09e3738fb3c93c585730f9546b7000fc32d018
Cardanoscan: https://preview.cardanoscan.io/transaction/7042a915046c87c8cd1d7b953f09e3738fb3c93c585730f9546b7000fc32d018
python3 src/off_chain/make_vest.py owner beneficiary  0.27s user 0.04s system 3% cpu 7.921 total
```

This is a speed up from 40s to 7s and the script only builds a simple transaction!

pycardano/backend/ogmios.py
```diff
@@ -54,12 +56,15 @@ def __init__(
         network: Network,
         compact_result=True,
         kupo_url=None,
+        refetch_chain_tip_interval=20,
```
cffls: We can avoid hardcoding this by using something like this `self.genesis_param.slot_length / self.genesis_param.active_slots_coefficient`

cffls (commented):
Nice! I can image how much queries this optimization will reduce :D

nielstron:
An optimal design would have a few threads that asynchronousmy fetch the chain tip, the protocol params, the epoch and such and fetch it. Then access to those values is instant and they are almost never outdated.

pycardano/backend/ogmios.py
nielstron: I would like to keep it as a parameter so that users can set it to 0, but agree we can default to these parameters

nielstron: done

cffls:
The test failed. I had a fix but can't push to your fork. Could you pull [the fix](https://github.com/Python-Cardano/pycardano/tree/feat/cache_chain_tip) into your fork?

nielstron:
Neat, multi-player PR

cffls (approved):

Issue #201, opened 2023-04-12
Querying UTxOs from an address never used before raises an exception


mateusap1:
If I create a new address and don't send any tokens to it, I would expect to get an empty list as a result if I used `chain_context.utxos(address)`. Currently, however, it raises the following exception:

```
      |   File "/usr/local/lib/python3.9/site-packages/blockfrost/utils.py", line 83, in recursive_append
      |     raise ApiError(request_response)
      | blockfrost.utils.ApiError: {'status_code': 404, 'error': 'Not Found', 'message': 'The requested component has not been found.'}
```

This is a snippet if you want to reproduce the error,

```python
import pycardano as pyc

skey = pyc.PaymentSigningKey.generate()
vkey = pyc.PaymentVerificationKey.from_signing_key(skey)

address = pyc.Address(payment_part=vkey.hash(), network=pyc.Network.TESNET)

chain_context = pyc.BlockfrostChainContext(...)

print(chain_context.utxos(str(address)))
```

This will give us the exception written above, but I would expect to get an empty list instead.

mateusap1:
The solution is to update the blockfrost chain context to try and catch the blockfrost api for this address. If it receives a 404 error, should return an empty list.

cffls:
This feature request makes sense. We may need to update Ogmios too.

cffls:
Closing this issue as it is addressed in #219

PR #219, opened 2023-04-19
Return empty utxo list from blockfrost when an address is never used before

cffls:


cffls:
This will fix https://github.com/Python-Cardano/pycardano/issues/201

PR #222, opened 2023-04-22
Include utxo with datum and script in utxo selection

cffls:
A utxo with datum and script attached isn't necessarily unspentable.

### DIFF
--- a/pycardano/txbuilder.py
+++ b/pycardano/txbuilder.py
@@ -1008,13 +1008,7 @@ class TransactionBuilder:
 
             for address in self.input_addresses:
                 for utxo in self.context.utxos(address):
-                    if (
-                        utxo not in seen_utxos
-                        and utxo not in self.excluded_inputs
-                        and not utxo.output.datum_hash  # UTxO with datum should be added by using `add_script_input`
-                        and not utxo.output.datum
-                        and not utxo.output.script
-                    ):
+                    if utxo not in seen_utxos and utxo not in self.excluded_inputs:
                         additional_utxo_pool.append(utxo)
                         additional_amount += utxo.output.amount
                         seen_utxos.add(utxo)


pycardano/txbuilder.py
```diff
@@ -993,9 +993,8 @@ def build(
                     if (
                         utxo not in selected_utxos
                         and utxo not in self.excluded_inputs
-                        and not utxo.output.datum_hash  # UTxO with datum should be added by using `add_script_input`
-                        and not utxo.output.datum
-                        and not utxo.output.script
+                        and not utxo.output.datum_hash  # UTxO with datum hash should be added by using
```
nielstron: Why are you keeping the hash check but removing datum and script?

nielstron (changes requested):

pycardano/txbuilder.py
nielstron: If the spending address is not a script, there is no need to provide the original datum to spend the UTxO :)

nielstron:
```suggestion
```

cffls: When a `datum_hash` is attached to a UTxO, it means that we only know the datum hash, but not the actual datum. In fact, we could only post datum hash before Vasil hardfork. To spend UTxOs with datum hash only, we need to provide the actual datum as part of the transaction. If we don't filter out these UTxOs, we can possibly build a tx that will never pass onchain validation.

nielstron: Thats not correct. Try it yourself :) If there is no PlutusScript validating the output, the datum does not matter.

Moreover you can still post datum hashes after the Vasil HF.

cffls: Yeah I think you are right. Datum is only required in phase 2 validation. If the address is not a script, datum doesn't matter. Thanks for correcting me.

Issue #204, opened 2023-04-14
FR: Potential inputs for TxBuilder


nielstron:
**Is your feature request related to a problem? Please describe.**
When interacting with multi-address wallets, there may be tons of addresses that a user wallet controls. Usually few of them hold UTxOs. Moreover browser wallets usually cache these UTxOs anyways and allow to retreive them quickly. 

When using pycardano to build a transaction for a browser wallet using TxBuilder, currently the only way for interaction is adding all of these addresses to the input_addresses. The TxBuilder will then fetch all addresses again - which may be very slow.

If you add all UTxOs of the wallet as inputs than necessarily all inputs will be consumed, which is also not necessarily desireable as it may make the transaction too big and decreases parallelizability of future tx builds (generating only one output)

**Describe the solution you'd like**
A parameter to the TxBuilder that allows passing a list of UTxOs that may be used by the coin selection algorithm, but do not need to be included in the tx as inputs.

**Describe alternatives you've considered**
None

**Additional context**
Useful when building dApps with PyCardano.

CIP30 defines a function that will return all UTxOs from the wallet: https://cips.cardano.org/cips/cip30/#apigetutxosamountcborvalueundefinedpaginatepaginateundefinedpromisetransactionunspentoutputnull

PR #223, opened 2023-04-22
Support potential inputs in txbuilder

cffls:
This feature will allow users to define a list of potential utxo, which could be returned by a wallet, for UTxO selector to select inputs from.
Addresses https://github.com/Python-Cardano/pycardano/issues/204

### DIFF
--- a/test/pycardano/test_txbuilder.py
+++ b/test/pycardano/test_txbuilder.py
@@ -1,3 +1,4 @@
+import copy
 from dataclasses import replace
 from test.pycardano.test_key import SK
 from test.pycardano.util import chain_context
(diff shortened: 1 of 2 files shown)


test/pycardano/test_txbuilder.py
```diff
@@ -122,6 +122,39 @@ def test_tx_builder_with_certain_input(chain_context):
...
 
 
+def test_tx_builder_with_potential_inputs(chain_context):
+    tx_builder = TransactionBuilder(chain_context, [RandomImproveMultiAsset([0, 0])])
+    sender = "addr_test1vrm9x2zsux7va6w892g38tvchnzahvcd9tykqf3ygnmwtaqyfg52x"
+    sender_address = Address.from_primitive(sender)
+
+    utxos = chain_context.utxos(sender)
+
+    tx_builder.potential_inputs.append(utxos[1])
```
nielstron: Should we test here that _not all_ inputs are chosen? This is the crucial property of potential inputs. I.e. maybe add two inputs such that the algorithm will never choose one of either because it does not contain the desired coin? Or choose a deterministic UTxO selection algo

nielstron (changes requested):
Perfect, this is exactly what I imagined. Ideally the test should also check the desired property of not generally adding all UTxOs.

test/pycardano/test_txbuilder.py
cffls: Yes, good point! Let me tweak this test a bit.

PR #225, opened 2023-04-24
Fix deep copy of plutus data

cffls:
copy.deepcopy doesn't work natively for CBORTag, which is frequently used in PlutusData. This commit implements `__deepcopy__` methods for PlutusData and RawPlutusData by serializing the object into cbor and deserializing them back to new objects.


juliusfrost (commented):
Tests look good to me!

PR #227, opened 2023-04-25
Freeze cbor tags as they are supposed to be hashable

nielstron:
This fixes this issue when trying to serialize the ScriptContext (which contains a CBorTag -> CBorTag map) 

https://github.com/OpShin/opshin-pioneer-program/actions/runs/4794620318/jobs/8528208767#step:10:167


nielstron:
I am actually not sure that this is the right approach as there may be cbortags with values != list. Ideally, all cbor primitives should be frozen maybe (including all lists, dicts, sets etc)?

PR #228, opened 2023-04-25
Freeze dictionary keys to make them hashable

nielstron:
Alternative solution for #227


nielstron:
This should fix https://github.com/OpShin/opshin-pioneer-program/pull/43

nielstron:
Only codecov failed uploading, possibly due to the CI being incorrectly set up

juliusfrost:
It looks like there's double recursion in _freeze and _helper, which squares the search depth. Maybe merge the two to freeze just the outer data type, since we can assume the inner values are already frozen.

cffls (approved):
This is awesome! 🚀

nielstron:
> It looks like there's double recursion in _freeze and _helper, which squares the search depth. Maybe merge the two to freeze just the outer data type, since we can assume the inner values are already frozen.

You're right, the Seth elements and dictionary keys do not need to be frozen again, they can be assumed to be frozen.

A frozenhelper function also sounds interesting 🤔

juliusfrost:
I'm looking at this again from my computer (as opposed to my phone), I think I meant to say merge `_dfs` with `_frozen`

juliusfrost:
@nielstron I made a quick edit here: https://github.com/OpShin/pycardano/pull/3

juliusfrost:
This might be a good opportunity to use hypothesis to generate arbitrary CBORSerializable types

juliusfrost:
If we don't want to freeze all the time we could set a freeze argument and only freeze when true, then set it true recursively for keys but default false otherwise

juliusfrost:
> If we don't want to freeze all the time we could set a freeze argument and only freeze when true, then set it true recursively for keys but default false otherwise

I implemented this here: https://github.com/OpShin/pycardano/pull/4

pycardano/serialization.py
```diff
@@ -214,46 +230,42 @@ def to_primitive(self) -> Primitive:
...
-                return new_result
+                    _dict[_dfs(k, freeze=True)] = _dfs(v, freeze)
+                if freeze:
+                    return frozendict(_dict)
+                return _dict
             elif isinstance(value, set):
-                return {_helper(v) for v in value}
-            elif isinstance(value, frozenset):
-                return frozenset({_helper(v) for v in value})
+                _set = set(_dfs(v, freeze) for v in value)
```
nielstron:
```suggestion
                _set = set(_dfs(v, freeze=True) for v in value)
```

juliusfrost:
https://github.com/OpShin/pycardano/pull/5
Fixes `TypeError: unhashable type: 'list'`

juliusfrost:
This recent [PR](https://github.com/Marco-Sulla/python-frozendict/pull/70) in frozendict fixes the mypy issue when I tested it locally. Hope they make a new release soon

juliusfrost:
https://github.com/OpShin/pycardano/pull/6 fixes issues with `poetry run mypy --install-types --non-interactive pycardano`

juliusfrost (approved):

juliusfrost:
@cffls This should be ready to merge :)

cffls:
Awesome, thanks! LGTM.

PR #253, opened 2023-06-08
Fix kupo plutusv2 script from kupo

### LINKED ISSUE #252: Reference input scripts are in wrong format
**Describe the bug**
looking at the bytes of a utxo.output.script I see b'\x01\x00\x002222222222222222222222222222222222222...' while the real script looks like this b'Y,\x1e\x01\x00\x0022222222222222222222222222222222222...'

PlutusV2Script(cbor2.loads(x)) == utxo.output.script passes, but it should be utxo.output.script == x --> one wrapping of cbor is missing. This is painful for later, where the hashing of the script fails and produces the wrong policyId for token minting

nielstron:
This should fix #252 but ideally there would be a test case for this


pycardano/backend/ogmios.py
```diff
@@ -343,13 +343,9 @@ def _utxos_kupo(self, address: str) -> List[UTxO]:
                     kupo_script_url = self._kupo_url + "/scripts/" + script_hash
                     script = requests.get(kupo_script_url).json()
                     if script["language"] == "plutus:v2":
-                        script = PlutusV2Script(
-                            cbor2.loads(bytes.fromhex(script["script"]))
-                        )
+                        script = PlutusV2Script(bytes.fromhex(script["script"]))
```
cffls, line 347: This might not work for the script that are double serialized. We should probably call `_try_fix_script` after this line like what we did in [blockfrost](https://github.com/Python-Cardano/pycardano/blob/main/pycardano/backend/blockfrost.py#L171).

cffls (commented):
Thanks for the fix! Suggested one improvement in the code.

pycardano/backend/ogmios.py
nielstron: Wait what? I would think that scripts are generally normalized on chain to either single or double serialization?

nielstron: Double/single serialization as I have faced it for now was just a difference in the format that frontends took the script, not how the node eventually interpreted it

cffls: Given the info in https://github.com/Python-Cardano/pycardano/pull/254, it seems to me double-wrapped script can exist on-chain (I also doubt that blockfrost would make such obvious frontend mistake). Since the [script checker](https://github.com/Python-Cardano/pycardano/blob/main/pycardano/backend/blockfrost.py#L38) verifies the script hash, it doesn't hurt to use it imo.

nielstron: Following the documentation in https://ogmios.dev/api/#schema-Tx%5BBabbage%5D the script hash involved is actually not stored in the transaction. So for ogmios, we can not apply this fix. I would assume though that the script returned is exactly the pre-image of the involved script-hash following @KtorZ s reasoning in https://github.com/Python-Cardano/pycardano/pull/254#issuecomment-1582122917

cffls: Thanks for updating the code! The build failed because import of `_try_fix_script` was missing.

Regarding #254 , do you mean ogmios is not able to fetch the script?

nielstron: Thanks, fixed it. Will continue the discussion about #254 in the corresponding PR :)

cffls (approved):

PR #254, opened 2023-06-08
Fix reference scripts fetched from ogmios

nielstron:
Same fix as for kupo (#253) for #252 

@ktorz maybe you can chime in on what exactly the format of scripts in utxos is in kupo/ogmios? Its mentioned to be[ "raw script"](https://cardanosolutions.github.io/kupo/#tag/Scripts/paths/~1scripts~1{script-hash}/get) from which I would actually understand 0 cbor wrapping but it appears to be exactly once?

### DIFF
--- a/pycardano/backend/ogmios.py
+++ b/pycardano/backend/ogmios.py
@@ -466,9 +466,9 @@ class OgmiosChainContext(ChainContext):
         script = output.get("script", None)
         if script:
             if "plutus:v2" in script:
-                script = PlutusV2Script(cbor2.loads(bytes.fromhex(script["plutus:v2"])))
+                script = PlutusV2Script(bytes.fromhex(script["plutus:v2"]))
             elif "plutus:v1" in script:
-                script = PlutusV1Script(cbor2.loads(bytes.fromhex(script["plutus:v1"])))
+                script = PlutusV1Script(bytes.fromhex(script["plutus:v1"]))
             else:
                 raise ValueError("Unknown plutus script type")
         datum_hash = (
--- a/pycardano/backend/ogmios.py
+++ b/pycardano/backend/ogmios.py
@@ -4,7 +4,6 @@ from datetime import datetime, timezone
 from enum import Enum
 from typing import Any, Dict, List, Optional, Tuple, Union
 
-import cbor2
 import requests
 import websocket
 from cachetools import Cache, LRUCache, TTLCache, func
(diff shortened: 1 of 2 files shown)


nielstron:
This fixed the issue I had in #252 in the tool I was using

KtorZ:
The scripts are provided in the format that correspond to the preimage of the various hashes calculated from scripts. For some unknown reasons, few interfaces like the cardano-cli have opted for encoding an already cbor-encoded binary object as cbor, resulting in this double-cbor-wrapping (dare I say it?) non-sense (albeit I have no doubt it wasn't intentional).  

Using only the "raw" flat-encoded binary form of the script is unsatisfactory IMO because it still requires being cbor-encoded before hashing to obtain a valid preimage. So the most meaningful form is the cbor-flat-encoded script. Now, why do we have 2 different binary encoding format, and one of them sometimes applied twice is a discussion for another day.

nielstron:
Thanks for this insight. So 

- if the binary is always singly-cbor wrapped i would like to ask for this being clarified in the kupo/ogmios docs
- if not, does this mean the preimage for the hash can be either singly or doubly wrapped? as discussed in https://github.com/Python-Cardano/pycardano/pull/253#discussion_r1222401917

pycardano/backend/ogmios.py
```diff
@@ -467,9 +467,9 @@ def _utxo_from_ogmios_result(self, result) -> UTxO:
         script = output.get("script", None)
         if script:
             if "plutus:v2" in script:
-                script = PlutusV2Script(cbor2.loads(bytes.fromhex(script["plutus:v2"])))
+                script = PlutusV2Script(bytes.fromhex(script["plutus:v2"]))
```
nielstron, line 468: It was suggested by @cffls to use `_try_fix_script` [here](https://github.com/Python-Cardano/pycardano/pull/253#discussion_r1225634371). However, ogmios does not return the script *hash* required to call the fixing method, so we can not know what the desired hash looks like.

nielstron:
@cffls bump, this needs a decision and IMO should be merged as is

nielstron:
Maybe we should add a test case for it? Ogmios is run in the test environment afaik

cffls:
> Maybe we should add a test case for it? Ogmios is run in the test environment afaik

Yes, we should add a test case for it before merging. I will try add one and see if it works.

cffls (approved):

PR #262, opened 2023-07-14
Implement new default constr ID

### LINKED ISSUE #239: Automatic unique CONSTR_IDs
**Is your feature request related to a problem? Please describe.**
When trying to define Union'd PlutusData in OpShin/PyCardano, one has to take care that constructor ids are unique for all elements in the Union. While other languages like PlutusTx take care of this automatically (assigning all elements in a Union unique ids at definition) in PyCardano we have to manually pick a set of unique CONSTR_IDs. Also we want to be able to benefit from the free declarability of Unions in Python, which lets us group together types across definitions (i.e Union[A,B] and Union[A,C], this is actually not possible with any other Cardano SC language !)

**Describe the solution you'd like**
We definetly want a few properties on the CONSTR_IDs
1) *small*: ideally the constr_id integer should be as small as possible, as smaller integers are encoded more efficiently in CBOR and save the end user minutxo and txfees (constr_ids are encoded as the cbor tag up to 7 bit size, after that encoded as generic integer)
1) *unique*: There should be as little overlap with other values as possible, so that we can group together classes in unions without having to worry about setting/overwriting the constr id
1) *overwritable*: In order to mimic external interfaces such as the ledger, we need to be able to manually set the CONSTR_ID. We also may want to manually set the IDs in case of collisions
1) *deterministic*: Datatypes that are defined in libraries may be imported in arbitrary contexts. the constr_id must therefore not depend on i.e. what other Unions the datatype is being used in or what other datatypes are declared in its surroundings

Overwritability is an implementation detail that we simply need to bear in mind when adopting this change.
IMO the best solution is based on a hash of the class name. The main question is now how to process the hash to serve for  the goal of minimal size.

##### Alternative A: `CONSTR_ID = hash(classname) % 127`

This solution truncates the hash of the classname to the last 7 bits, giving us a constructor id between 0 and 127, neatly encodable in a single byte. However, due to the birthday paradox, this may result in a collision with another constructor id from 23 defined classes onwards. In practice we see ~5-10 user defined classes per contract so this could be fine if alternative B is too expensive

##### Alternative B: `CONSTR_ID = hash(classname) % (2**32)`

This solution truncates the hash of the classname to 32 bits, nicely fitting into a normal integer. This will reduce the likelyhood of conflicts to a negligible amount, however costing 3 bytes more than solution A per datum (mind that these datums end up on chain). It should be investigated how much the impact of this is on minutxo and script execution cost.

**Describe alternatives you've considered**

Solutions similar to `auto()` in enum (global incremental counter that increases with every definition of a plutusdatum) is not a good idea because it breaks determinism.

**Additional context**
[OpShin](opshin.dev/) lets you write smart contracts in Cardano and uses the PlutusData class for native data.
In haskell, data is usually defined together with all alternatives of the same type (i.e. `A = B | C | D`) while in python/Pycardano/opshin the classes are first defined, then later combined to unions (i.e. `class B; class C; class D; A = Union[B, C, D]`)

Requesting comments from @cffls and @juliusfrost, but will most likely just go ahead and implement this as soon as I got time for it (~1-2 weeks from now)

nielstron:
This is a possible implementation of a fix for #239 

Properties

- almost unique
   - pretty much random value between 1-2**32
   - based on class name, field names and field types
- deterministic (sha256 defined behaviour)
- overwritable (just an attribute)

### DIFF
--- a/pycardano/plutus.py
+++ b/pycardano/plutus.py
@@ -7,7 +7,7 @@ import json
 from dataclasses import dataclass, field, fields
 from enum import Enum
 from hashlib import sha256
-from typing import Any, Optional, Type, Union
+from typing import Any, ClassVar, Optional, Type, Union
 
 import cbor2
 from cbor2 import CBORTag
@@ -468,6 +468,9 @@ class PlutusData(ArrayCBORSerializable):
         >>> assert test == Test.from_cbor("d87a9f187b43333231ff")
     """
 
+    AUTO_ID: ClassVar[bool] = False
+    """Enables automatic assignment of a deterministic constructor id (CONSTR_ID) for the Plutus data if set to True."""
+
     @classproperty
     def CONSTR_ID(cls):
         """
@@ -476,6 +479,9 @@ class PlutusData(ArrayCBORSerializable):
         The default implementation is an almost unique, deterministic constructor ID in the range 1 - 2^32 based
         on class attributes, types and class name.
         """
+        if not cls.AUTO_ID:
+            return 0
+
         k = f"_CONSTR_ID_{cls.__name__}"
         if not hasattr(cls, k):
             det_string = (
--- a/pycardano/plutus.py
+++ b/pycardano/plutus.py
@@ -837,3 +837,8 @@ def script_hash(script: ScriptType) -> ScriptHash:
         )
     else:
         raise TypeError(f"Unexpected script type: {type(script)}")
+
+@dataclass
+class Unit(PlutusData):
+    """ The default "Unit type" with a 0 constructor ID """
+    CONSTR_ID = 0
--- a/pycardano/plutus.py
+++ b/pycardano/plutus.py
@@ -476,16 +476,17 @@ class PlutusData(ArrayCBORSerializable):
         The default implementation is an almost unique, deterministic constructor ID in the range 1 - 2^32 based
         on class attributes, types and class name.
         """
-        if not hasattr(cls, "_CONSTR_ID"):
+        k = f"_CONSTR_ID_{cls.__name__}"
+        if not hasattr(cls, k):
             det_string = (
                 cls.__name__
                 + "*"
                 + "*".join([f"{f.name}~{f.type}" for f in fields(cls)])
             )
             det_hash = sha256(det_string.encode("utf8")).hexdigest()
-            setattr(cls, "_CONSTR_ID", int(det_hash, 16) % 2**32)
+            setattr(cls, k, int(det_hash, 16) % 2**32)
 
-        return cls._CONSTR_ID
+        return getattr(cls, k)
 
     def __post_init__(self):
         valid_types = (PlutusData, dict, IndefiniteList, int, bytes)
--- a/pycardano/plutus.py
+++ b/pycardano/plutus.py
@@ -838,7 +838,9 @@ def script_hash(script: ScriptType) -> ScriptHash:
     else:
         raise TypeError(f"Unexpected script type: {type(script)}")
 
+
 @dataclass
 class Unit(PlutusData):
-    """ The default "Unit type" with a 0 constructor ID """
+    """The default "Unit type" with a 0 constructor ID"""
+
     CONSTR_ID = 0
--- a/pycardano/plutus.py
+++ b/pycardano/plutus.py
@@ -7,7 +7,7 @@ import json
 from dataclasses import dataclass, field, fields
 from enum import Enum
 from hashlib import sha256
-from typing import Any, ClassVar, Optional, Type, Union
+from typing import Any, Optional, Type, Union
 
 import cbor2
 from cbor2 import CBORTag
@@ -468,9 +468,6 @@ class PlutusData(ArrayCBORSerializable):
         >>> assert test == Test.from_cbor("d87a9f187b43333231ff")
     """
 
-    AUTO_ID: ClassVar[bool] = False
-    """Enables automatic assignment of a deterministic constructor id (CONSTR_ID) for the Plutus data if set to True."""
-
     @classproperty
     def CONSTR_ID(cls):
         """
@@ -479,9 +476,6 @@ class PlutusData(ArrayCBORSerializable):
         The default implementation is an almost unique, deterministic constructor ID in the range 1 - 2^32 based
         on class attributes, types and class name.
         """
-        if not cls.AUTO_ID:
-            return 0
-
         k = f"_CONSTR_ID_{cls.__name__}"
         if not hasattr(cls, k):
             det_string = (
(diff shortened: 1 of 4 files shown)


nielstron:
As suggested by @jkoppel we should maybe use constructor IDs up to 2^32 since the numbers up to this value are treated all the same by the plutus VM wrt costing. I suspect that the additional 3 bytes don't make much of a costing difference either when looking at script and inline datum size.

pycardano/plutus.py
```diff
@@ -460,9 +468,19 @@ class will reduce the complexity of serialization and deserialization tremendous
...
+    @classproperty
+    def CONSTR_ID(cls):
+        """
+        Constructor ID of this plutus data.
+        It is primarily used by Plutus core to reconstruct a data structure from serialized CBOR bytes.
+        The default implementation is an almost unique, deterministic constructor ID in the range 1 - 2^32 based
+        on class attributes, types and class name.
+        """
+        det_string = (
+            cls.__name__ + "*" + "*".join([f"{f.name}~{f.type}" for f in fields(cls)])
```
cffls: Does it make sense to sort the fields before hashing? 

Personally, I'd like to see both 

```python
    @dataclass
    class A(PlutusData):
        a: int
        b: bytes
```

and 

```python
    @dataclass
    class A(PlutusData):
        b: bytes
        a: int
```

hash into the same digest.

cffls: Also, is there a way to cache this result? The calculation might be expensive.

cffls (commented):
Elegant implementation!

pycardano/plutus.py
nielstron: > Does it make sense to sort the fields before hashing?

No this doesnt make sense. The order of fields is relevant for plutusData! (think of how C objects have their fields arranged in memory, same applies to Plutus data how their fields are arranged in the internal list). In particular, the two classes you listed do not constitute the same plutus data

cffls: Okay makes sense, thanks for the explanation!

nielstron: I played around with some trivial caching methods (i.e. overriding CONSTR_ID after initial call) but get some really weird errors in pytest (CONSTR_IDs are the same for different classes, but only if run in the suite, not when run individually)

cffls: > I played around with some trivial caching methods (i.e. overriding CONSTR_ID after initial call) but get some really weird errors in pytest (CONSTR_IDs are the same for different classes, but only if run in the suite, not when run individually)

Will take a look at this issue.

cffls: This seems to work:

```python
    @classproperty
    def CONSTR_ID(cls):
        """
        Constructor ID of this plutus data.
        It is primarily used by Plutus core to reconstruct a data structure from serialized CBOR bytes.
        The default implementation is an almost unique, deterministic constructor ID in the range 1 - 2^32 based
        on class attributes, types and class name.
        """
        if not hasattr(cls, "_CONSTR_ID"):
            det_string = (
                cls.__name__ + "*" + "*".join([f"{f.name}~{f.type}" for f in fields(cls)])
            )
            det_hash = sha256(det_string.encode("utf8")).hexdigest()
            setattr(cls, "_CONSTR_ID", int(det_hash, 16) % 2**32)

        return cls._CONSTR_ID
```

If you could grant me write access to `https://github.com/OpShin/pycardano.git`, I can push this change. Also, I think there is an option to allow others to push changes to your PR when creating it.

nielstron: Cool! I have given you write access

pycardano/plutus.py
```diff
@@ -460,9 +470,22 @@ class will reduce the complexity of serialization and deserialization tremendous
...
+    def CONSTR_ID(cls):
+        """
+        Constructor ID of this plutus data.
+        It is primarily used by Plutus core to reconstruct a data structure from serialized CBOR bytes.
+        The default implementation is an almost unique, deterministic constructor ID in the range 1 - 2^32 based
+        on class attributes, types and class name.
+        """
+        if not hasattr(cls, "_CONSTR_ID"):
+            det_string = (
+                cls.__name__ + "*" + "*".join([f"{f.name}~{f.type}" for f in fields(cls)])
```
cffls: Some fields are metadata, e.g. ClassVar, which is not serialized to cbor. I think we should try to exclude those fields.

nielstron: Maybe you can recycle the logic from the to_cbor or to_primitive function? these would exactly describe which parts are not serialized if I understand correctly?

pycardano/plutus.py
```diff
@@ -460,9 +468,31 @@ class will reduce the complexity of serialization and deserialization tremendous
         >>> assert test == Test.from_cbor("d87a9f187b43333231ff")
     """
 
-    CONSTR_ID: ClassVar[int] = 0
-    """Constructor ID of this plutus data.
-       It is primarily used by Plutus core to reconstruct a data structure from serialized CBOR bytes."""
+    AUTO_ID: ClassVar[bool] = False
```
nielstron: I would honestly much prefer if AUTO ID would default to true.

When writing a smart contract and defining classes, setting the constr id manually requires more expertise than automatically having the unique constructor. If you think about it, there is no reason you should even know about constructor IDs when writing smart contracts, you should be able to assume that classes are distinct and distinguishable. So setting the constructor manually should be the non-default, reserved for experts that know what they are doing.

When you are using pycardano to model some existing contract written in Plutus, you definitely should be an expert and manually specify the constructor id, as you are working across different language implementations.

This is a breaking change, but I think it would much benefit the usability of pycardano.

Making AUTO ID default to True would then make the entire flag superfluous - when you are setting it to false manually you can just as well manually specify the constructor id.

cffls: Makes sense to enable auto id by default as you explained. Added auto id because I wanted to make it a non-breaking change. I will revert this commit.

pycardano/plutus.py
cffls: It's a bit tricky to reuse to_cbor or to_primitive here because they are not class methods. We can check if the type is in [primitive list](https://github.com/Python-Cardano/pycardano/blob/main/pycardano/serialization.py#L70) (recursively).

nielstron:
@cffls any update on this?

pycardano/plutus.py
cffls: As of now, all fields are checked and serialized. We can revisit this when field exclusion is enabled in the future.

cffls:
Added a comment regarding field exclusion. The PR looks good to me. If you don't have more to add, I will merge it.

nielstron:
It looks good to me as well :)

cffls (approved):

PR #264, opened 2023-07-20
Fix: negative minted amount is also requested

nielstron:
For transactions that only specified a token to be burned it could currently happen that the token was not selected by the input selector - because negative minting amounts were not counted as requested by the tx builder. This is a bugfix.

- [x] Add a unit test

### DIFF
--- a/pycardano/txbuilder.py
+++ b/pycardano/txbuilder.py
@@ -934,7 +934,13 @@ class TransactionBuilder:
             selected_amount += i.output.amount
 
         if self.mint:
-            selected_amount.multi_asset += self.mint
+            # Add positive minted amounts to the selected amount (=source)
+            for pid, m in self.mint.items():
+                for tkn, am in m.items():
+                    if am < 0:
+                        selected_amount += Value(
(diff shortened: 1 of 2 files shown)


MuesliSwapTeam:
The only error appears to have been uploading to Codecov

nielstron:
The only error appears to have been uploading to Codecov

nielstron:
Note: this breaks if the burned amount is manually added as input, because the minted amount is added to the `selected_amount`, resulting in `0`. This is valid as the amount can not be used anymore later. However modeling it as a requested output leverages the coinselection to automatically fetch the required input for the burning.

cffls:
Nice! I will try to add a test. Btw, is there a way to allow me to push to your branch ([OpShin:fix/minted_amount_requested](https://github.com/OpShin/pycardano/tree/fix/minted_amount_requested))?

nielstron:
You should have access now :)

nielstron:
I fixed the previously mentioned issue when pre-selected inputs contain the token to be burned. What's missing is a unit test that checks the automatic selection of tokens to be burned

cffls (approved):

PR #269, opened 2023-09-21
bugfix: fixed incorrect bytestring encoding PlutusData

### LINKED ISSUE #268: Recreating long submission bytestream datum
**Describe the bug**
I am trying to submit to a contract. One of the `bytes` in the datum is going to be more than 64 bytes long. Every time I submit to blockfrost, I get an error back saying that the bytestring cannot be longer than 64 bytes. I narrowed it down to the bytestring that is more than 64 bytes long.

I have even gone so far as to copy the hexstring from the application of interest, ingest with pycardano, sign, and submit and I get the same error. I am certain is not a configuration error on my part, unless the transaction is being parsed incorrectly on ingestion.

I guess I'm wondering what the issue could be here. Blockfrost? Is the application using an invalid bytestring length? If so, how are they able to submit this? Could Pycardano be sending the transaction incorrectly to Blockfrost?

**Logs**
Here is a reference transaction:
https://preprod.cardanoscan.io/transaction/a165c606d2600324b51b15d138bb91eb6bef46e33d51eaebbc7d78e0702f12c1

And the associated datum:
https://preprod.cardanoscan.io/datumInspector?datum=d8799f581c100bba42156945783bfc45ad9b0f9969a5fc34d186914cd528f439bf581cb92203c0c15c0b0da3e7ac99d48de85cad1beda530b8f54a12e9e33b584024ccdb90a49a970539327b45d585624d027f1d4c05c521a2c9cbe805d3d02cb74c08f25d32f5cc923e22badd041902424d5dbe53886f349adac7912b89a095075f584066756e64696e672d622758205c7831643365715c7864625c786663635c7839322c5c783837735c783865795c7864324a5c7839665c7865345c7864355c78656158295c7831393a5c783935365c7839665c786662745c7831665c7863645c7831335c7831622e5c78633127ff1a02747cd00a01ff

theeldermillenial:
Fix #268 

## Summary

`pycardano` incorrectly encodes `bytes` longer than 64 bytes for `PlutusData`. Currently, a `bytes` element is encoded the same regardless of length, but if the length is larger than 64 bytes it should be broken up into 64 byte chunks per the following spec:
https://developers.cardano.org/docs/get-started/cardano-serialization-lib/transaction-metadata/#metadata-limitations

This PR fixes this issue by creating a dummy class to catch `PlutusData` objects during serialization and properly breaks up `bytes` values longer than 64 bytes in length.

If further explanation is needed, I can provide examples. This PR provides a unit test to verify the expected output is obtained with a long `bytes` input.


nielstron:
Is there a reason to introduce MetadataIndefiniteList? I don't see the benefit over using IndefiniteList directly

pycardano/serialization.py
```diff
@@ -166,17 +173,31 @@ def default_encoder(
...
+        value, (IndefiniteList, IndefiniteFrozenList, MetadataIndefiniteList)
+    ):
         # Currently, cbor2 doesn't support indefinite list, therefore we need special
         # handling here to explicitly write header (b'\x9f'), each body item, and footer (b'\xff') to
         # the output bytestring.
         encoder.write(b"\x9f")
         for item in value:
-            encoder.encode(item)
+            if (
+                isinstance(value, MetadataIndefiniteList)
```
nielstron: I don't see why the encoding as MetadatIndefiniteList is necessary, can we not simply encode all bytes longer than 64 bytes as a list?

theeldermillenial: That's a good question. I'm still new to Cardano. I guess these were my thoughts and how I would push back on encoding all bytes the same way.

1. If bytes length only has a restriction for metadata, will encoding them the same way in other parts of the message cause an issue?
2. Why impose the same criteria on other parts of the message if it's not strictly required?
3. Won't chunking bytes data cause nominally larger message lengths, and by extension, marginally higher tx fees?

If you feel these are non-issues, I think it's easy enough to remove the dummy class and encode everything the same way.

theeldermillenial: I think in retrospect your comments on the issue I raised make more sense now.

nielstron: IIRC the cardano ledger generally specifies that the cbor encoding of bytestrings should be at most 64 bytes long each piece. This would prevent OOM attacks when reading very long bytestrings. However the ledger does not enforce this, leading to different implementations being abound on chain.

theeldermillenial: But isn't OOM attack prevented my maximum transaction size anyway? Again, just playing devils advocate here. I'm still relatively new to Cardano.

I can revert changes back to the original without the dummy class for `PlutusData`. Just give me a yay or nay. With these changes I was able to successfully submit to smart contracts, so I know these changes work correctly.

theeldermillenial: Just wanted to bump this so I can finish it off :)

nielstron: I would prefer this without the dummy class - maybe you can revert to that and see if both the test cases and your submission pass?

then Jerry or I can create a test case for exactly this datum submission

theeldermillenial:
Removed the dummy class.

All unit tests pass, including the specific test created to test a long bytestring.

My specific use case works.

I'm happy to do any additional work required to finish this off. I didn't see any contributing doc, so I don't know how versioning is handled.

pycardano/serialization.py
```diff
@@ -176,7 +176,14 @@ def default_encoder(
         # the output bytestring.
         encoder.write(b"\x9f")
         for item in value:
-            encoder.encode(item)
+            if isinstance(item, bytes) and len(item) > 64:
+                encoder.write(b"\x5f")
```
cffls: This is only activated when an item is inside a indefinite list. Do we need to break byte strings that are not part of indefinite list?

cffls (commented):
Thanks for fixing this! Posted a question in the code. Apology for the delayed response.

pycardano/serialization.py
nielstron: AFAIK we need to break all bytes that are longer than 64 bytes

theeldermillenial: I may have misunderstood, but it seemed to me that this was the best place to put it since all `PlutusData` are cast to `IndefiniteList`.

If I pull it out of the `IndefiniteList` block, will it be handled properly? I guess it should.

nielstron: I guess you (correctly) noticed that all PlutusData _fields_ are part of an indefinite list. However plutusdata can also contain bytes without being part of PlutusData (i.e. pure bytes or bytes that are keys in dictionaries)

theeldermillenial: So is the final answer to pull it outside of the `IndefiniteList` block?

nielstron: yes :)

theeldermillenial: I like to make sure things are explicit :)

So, I just tried pulling the `bytes` encoding outside of the `IndefiniteList` block, but it doesn't appear as though raw types ever make it to the `default_encoder`. Do I need to modify something else somewhere?

nielstron: This seems to be indeed the case. I will check with the cbor2 documentation if the default behaviour of the cbor2 encoder can be changed (so far it only allows _additional_ rules for unknown objects. We may need to wrap bytes into a new type rather than the lists :)

theeldermillenial: so...back to dummy classes? 😂

I am not annoyed at all. I just appreciate the development process :)

Let me know what you find out. I can also take a look into it, but I'm less familiar with the cbor2 library.

nielstron: In this [documentation](https://cbor2.readthedocs.io/en/latest/customizing.html) it seems that yes, we need dummy classes. But not for lists, for bytes! :) 

I am also wondering if there are cases where integers are incorrectly encoded (when they exceed 64 bytes size) since I implemented a special case for this here: https://github.com/OpShin/uplc/blob/448f634cc1225de6dd7390b670b01396d2e71156/uplc/ast.py#L430

theeldermillenial: I guess I am seeing more and more the intuition behind all the custom classes in OpShin.

I realize it's a bigger lift, but is there any reason why we wouldn't just take OpShin's implementation and pull it over to here? Then, just rely on pycardano rather than duplicating efforts across repos?

I apologize if I'm speaking out of ignorance and there are things I'm not considering, but this seems like it might be the more lasting implementation.

nielstron: No worries at all. The code I wrote for OpShin/UPLC was created after pycardano was written, hence there might be a point in copying it over. Then again, the UPLC implementation is really only catered towards PlutusData, while PyCardano also handles serialization of all other kinds of things - not sure if anything will break.

Long story short: The only reason that there are two different implementations is that no one yet tried to unify them.

theeldermillenial: Okay, I would like to have this done sooner rather than later. Can I just create a dummy class for bytes to patch this and open a more general issue about syncing datum handling between OpShin and pycardano?

nielstron: Yes sounds good to me! Would also prefer to get this resolved over any big open stale PR :)

theeldermillenial:
Alright, this was annoying. I made a new dummy class to wrap `bytes`. That seemed to properly wrap the `PlutusData` tests so that the unit tests pass (with a modification to the ground truth CBOR for one test), but wrecked about 20 other unit tests that did raw dictionary/list comparisons to primitives. I was able to get around that by modifying the comparator for the dummy class.

There's still one failing unit test that has to do with generating the script hash. I didn't have a chance to dig into it. Maybe you can take a look and more easily see what might be the issues. Once that is resolved, this should be good.

theeldermillenial:
Alright, as best as I can tell, the last failing test was caused by serializing the `COST_MODEL`. If dictionaries are serialized to `bytes`, then the hash should change with the new way data is serialized to cbor.

test/pycardano/test_util.py
```diff
@@ -149,7 +149,7 @@ def test_script_data_hash():
     redeemers = [Redeemer(unit, ExecutionUnits(1000000, 1000000))]
     redeemers[0].tag = RedeemerTag.SPEND
     assert ScriptDataHash.from_primitive(
-        "032d812ee0731af78fe4ec67e4d30d16313c09e6fb675af28f825797e8b5621d"
+        "b11ed6f6046df925b6409b850ac54a829cd1e7603145c9aaf765885d8ec64da7"
```
cffls: Not sure if this should change. If we use write the same test in Haskell, it would generate the same hash.

pycardano/serialization.py
```diff
@@ -240,6 +265,8 @@ def to_primitive(self) -> Primitive:
         def _dfs(value, freeze=False):
             if isinstance(value, CBORSerializable):
                 return _dfs(value.to_primitive(), freeze)
+            elif isinstance(value, bytes):
+                return ByteString(value)
```
cffls: IMO, it seems incorrect to replace every bytes with ByteString. Instead, we just offer ByteString for users to use in PlutusData or Metadata. 
For some internal implementation that generates bytes as intermediate values, e.g. [script_data_hash](https://github.com/Python-Cardano/pycardano/blob/eaadc385e10f94355d1e632e3c75c9c3f5e3e46c/pycardano/utils.py#L212), we don't want to change its type arbitrarily.

nielstron: Should be implementable by a simple parameter?

theeldermillenial: That was one of my original points and how I originally had it implemented. 

Can someone please make a definitive final decision so I can fix and be done? I've implemented and reimplemented this multiple times.

test/pycardano/test_util.py
theeldermillenial: I think as you noticed in the other comment, this changes because all bytes are being encoded the same way per nielstrons suggestion.

Your comment makes sense. If we only change encoding in metadata/plutusdata, then the hash would not change.

pycardano/serialization.py
cffls: @nielstron Could you elaborate how adding a parameter will work?

nielstron: It might not work as straightforward as I imagined it. @theeldermillenial was right, maybe we should just roll with the initial design. I appreciate the excourse though because now we know precisely which bytes to encode this way 😅 sorry for the divergence.


maybe we can document this (and ideally find some supporting documentation on the discrepancy in the implementation)

theeldermillenial: You just earned my respect with that. I appreciate it.

So here's what I propose. We create a variation on the original idea of a dummy class for `PlutusData`. Probably use that for metadata as well?

I would prefer that we try to get it right this time so we can finish it off. If someone wants to tackle it, that's okay. I'm happy to work on it, but I want to hear a consensus before starting to work on it this time around.

cffls: My preferred approach is to offer users `ByteString` class to use, which the encoder can automatically break it down to byte array. If a bytes object is found longer than 64 instead, pycardano should raise an exception and recommend users to use `ByteString`.

theeldermillenial: @cffls But should this error only be thrown for `PolusData` and Metadata? Or should we apply it globally?

My two cents is "only implement exactly what is defined". The bytes length restriction appears to be limited to "metadata", so maybe we only apply it to metadata.

cffls: We already have this check for metadata: https://github.com/Python-Cardano/pycardano/blob/main/pycardano/metadata.py#L40-L49

I thought this should be also enforced in PlutusData, which was the reason why this PR was raised. If not, I am fine with only providing ByteString as an option for user in this PR.

theeldermillenial: My apologies. I misspoke. When I said Metadata, I also meant PlutusData.

Also, I now see exactly what you're saying, and I think you're solution makes the most sense. You are saying we should inject a check for long byte strings in PlutusData and throw an error similar to what is seen in Metadata. Part of that error message should indicate the user can use the new ByteString class to allow longer bytes.

I think this is the most transparent approach, and it keeps in line with what I see to be pycardano's philosophy of being very unbiased.

If this is what you mean, I'll make the changes and we can be done. I will revert any hashes I altered, since this should really only affect the test I created. If there's any other unit tests you would like to see, I'm happy to add them.

cffls: Yes, this is exactly what I meant. Please go ahead with this approach. Thank you for confirming. ☺️

cffls (commented):
Looks good to me. Some code failed mypy analysis, you can do `make qa` to check errors.

theeldermillenial:
Should be good to go now.

Now that I know all your qa stuff, it should be easier for me to contribute. I'm dedicated to building stuff on Cardano in Python, so I'll try to contribute as I find things. Like #273

cffls (approved):
LGTM. Thank you for your contribution!

PR #272, opened 2023-10-04
Reconstructable deterministic constrids

### LINKED ISSUE #271: Unique constructor id can not be independently implemented in foreign languages
The computation of the unique constructor id is currently dependent on the stringified representation of the type of PlutusData attributes. This makes much sense for primitives and container types. However, the type may also be another Plutus Data class - this may lead to the problem that the stringified representation of such a class is very difficult to predict from the perspective of another language. The implementation should stay agnostic to the actual Python implementation though.

There are two ways to mitigate this: instead of the stringification of the type, we use
- the stringification of the type structure (recursively resolved to primitive types)
- the constructor id of the type

The constructor id however may not be unique in all cases so I would vouch for the first approach.

References:
https://github.com/Python-Cardano/pycardano/blob/0a95536e9e3779dcfc90eda7fcfa23e389a79c3f/pycardano/plutus.py#L484C43-L484C47

I am willing to work on a reference solution to fix this

nielstron:
This resolves #271 

Please reach out if there are any remaining questions as to the necessity or implementation of this.


pycardano/plutus.py
```diff
@@ -447,6 +448,45 @@ def get_tag(constr_id: int) -> Optional[int]:
...
+        return (
+            "cons["
+            + cls.__name__
+            + "]("
+            + (str(cls.CONSTR_ID) if not skip_constructor else "_")
+            + ";"
+            + ",".join(f.name + ":" + id_map(f.type) for f in fields(cls))
+            + ")"
+        )
+    raise TypeError(f"Unexpected type for automatic constructor generation: {cls}")
```
cffls: This is nice. Is there a language-agnostic documentation for it?

cffls (commented):
A small conflict otherwise lgtm.

pycardano/plutus.py
nielstron: No. Basically, this  is the reference implementation for this kind of constructor IDs. I could write up a documentation, either in this document (where?), OpShin or in a dedicated CIP

cffls: I will leave you to decide. For me, because this can potentially get adopted by other languages, I think a dedicated CIP is the best choice.

nielstron: I have added a CIP here: https://github.com/cardano-foundation/CIPs/pull/608

cffls:
Tried merging from main branch, but failed with this error:

```
FAILED test/pycardano/test_plutus.py::test_plutus_data_long_bytes - TypeError: Unexpected type for automatic constructor generation: <class 'pycardano.serialization.Byt...
```

Looks like `ByteString` introduced in this [PR](https://github.com/Python-Cardano/pycardano/pull/269) is not supported by id map yet. Do you think we should add specification for `ByteString` type, or convert it to list of bytes when generating constructor id?

nielstron:
I have added support for `ByteString` now, it should behave the same as bytes (as the existance of ByteString is just an implementation detail of pycardano and irrelevant to third party implementations)

cffls (approved):
LGTM, thanks!

PR #273, opened 2023-10-06
Added extended signing key support for cip8

### LINKED ISSUE #270: CIP8 signing with extended signing key fails
**Describe the bug**
When using an extended signing key, either payment or staking, signing a message using `cip8.sign` produces an error.

**To Reproduce**
Modify the `test_cipy8.py` tests to use extended signing keys from the `test_keys.py` file. Run the test.

**Logs**
The error generated reads as follows:
```bash
pycardano/cip/cip8.py:88: in sign
    encoded = msg.encode()
.venv/lib/python3.10/site-packages/cose/messages/sign1message.py:67: in encode
    message = [self.phdr_encoded, self.uhdr_encoded, self.payload, self.compute_signature()]
.venv/lib/python3.10/site-packages/cose/messages/signcommon.py:63: in compute_signature
    return alg.sign(key=self.key, data=self._sig_structure)
.venv/lib/python3.10/site-packages/cose/algorithms.py:1021: in sign
    sk = Ed25519PrivateKey.from_private_bytes(key.d)
.venv/lib/python3.10/site-packages/cryptography/hazmat/primitives/asymmetric/ed25519.py:77: in from_private_bytes
    return backend.ed25519_load_private_bytes(data)
```

**Expected behavior**
Since `cip8.sign` checks to see if one of the keys is a `StakeExtendedSignKey`, I would expect that at least a `StakeExtendedSignKey` to work correctly.

theeldermillenial:
Fixes #270 

This PR is an attempt at implementing extended signing key support for cip8 message signing and verification.

I am not confident that this is the correct approach, but it has the right pieces in place.

It wasn't clear to me whether I should include the public key plus the chain code for the extended verification key. It seemed obvious to me that you would, but I haven't played around with this. I did created a unit test, but it's not in depth.

I do have an application I can test against, but I thought I would open for comment.

### DIFF
--- a/pycardano/cip/cip8.py
+++ b/pycardano/cip/cip8.py
@@ -21,7 +21,6 @@ from pycardano.key import (
     StakeExtendedSigningKey,
     StakeSigningKey,
     StakeVerificationKey,
-    VerificationKey,
 )
 from pycardano.network import Network
 
--- a/pycardano/cip/cip8.py
+++ b/pycardano/cip/cip8.py
@@ -1,7 +1,6 @@
 from typing import Optional, Union
 
 from cbor2 import CBORTag, dumps
-
 from cose.algorithms import EdDSA
 from cose.headers import KID, Algorithm
 from cose.keys import CoseKey
@@ -14,10 +13,10 @@ from cose.messages import CoseMessage, Sign1Message
 from pycardano.address import Address
 from pycardano.crypto import BIP32ED25519PublicKey
 from pycardano.key import (
-    PaymentVerificationKey,
-    SigningKey,
     ExtendedSigningKey,
     ExtendedVerificationKey,
+    PaymentVerificationKey,
+    SigningKey,
     StakeExtendedSigningKey,
     StakeSigningKey,
     StakeVerificationKey,
@@ -129,7 +128,6 @@ def sign(
 def verify(
     signed_message: Union[str, dict],
     attach_cose_key: Optional[bool] = None,
-    extended=False,
 ) -> dict:
     """Verify the signature of a COSESign1 message and decode its contents following CIP-0008.
     Supports messages signed by browser wallets or `Message.sign()`.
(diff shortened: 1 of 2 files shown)


theeldermillenial:
I didn't see the `.flake8` config in there, or I would have run it before opening the PR. Apologies.

Should be good now.

theeldermillenial:
After some testing, I'm not certain this is the appropriate solution. However, I don't know what is. I think ideally we could extract the non extended payment/staking signing and verification keys from the extended keys. That would simplify this and make it consistent.

If someone could provide some guidance, I can work on implementing.

theeldermillenial:
Actually, this does work! When I was doing testing I used the wrong test wallet to validate.

My validation test was to sign a message to get a validation key for Iagon, then perform a bunch of API calls to create/delete files and folders.

Then I logged into the dashboard, which requires a CIP8 signed message (signed with Eternl).

When I signed the message for Iagon, it successfully validated and let me perform API calls for my address.

I think details about the hackiness of the solution may need to be worked out. I basically pulled apart the internals of the COSE library to achieve this. Maybe there is a better solution here.

cffls (approved):
Looks good to me. I don't think the solution is hacky. 😁

theeldermillenial:
I guess one final question I would have before merging is whether we can autodetect whether the signing was done with an extended key. Right now I added an `extended` parameter to `verify`. I forget if this is something you can detect based on the number of bytes in the verification key.

If this is a needed parameter, lets send it! I'm happy I've been able to contribute back.

pycardano/cip/cip8.py
```diff
@@ -175,7 +196,16 @@ def verify(
     # attach the key to the decoded message
     decoded_message.key = cose_key
 
-    signature_verified = decoded_message.verify_signature()
+    if extended:
+        vk = BIP32ED25519PublicKey(
+            public_key=verification_key[:32], chain_code=verification_key[32:]
+        )
```
cffls, line 200: You are right, we can probably determine whether the key is an extended key by simply checking its length. 64 bytes means it is an extended key, while 32 bytes means non-extended key.

astrojarred:
Just saw this PR-- awesome work @theeldermillenial :)

theeldermillenial:
Sorry that took a few days.

Should be good to go now.

What kind of a release schedule do you have with this? I have two different packages that are dependent on the different fixes I've implemented :)

No rush, just curious.

cffls (approved):

PR #293, opened 2024-01-10
Fix plutus script deserialization

### LINKED ISSUE #251: Cannot deserialize PlutusV2 script attached to transaction
**Describe the bug**
Trying to deserialize a transaction from cborhex that was generated and hexlified by PyCardano before, I get the following error
```
Jun 07 23:45:25 ********[12055]: pycardano.exception.DeserializeException: Cannot deserialize object:
Jun 07 23:45:25 ********[12055]: [b'Y,\x1e\x01\x00\x002222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222>
Jun 07 23:45:25 *********[12055]:  in any valid type from (typing.List[pycardano.plutus.PlutusV2Script], <class 'NoneType'>).
```

**To Reproduce**
Serialize a Tx with a PlutusV2 script attached, then deserialize again.

**Logs**
Logs are above

**Expected behavior**
Transaction.from_cbor(tx.to_cbor()) should work without problems

**Environment and software version (please complete the following information):**
 - OS: Ubuntu 22
 - PyCardano Version: latest

**Additional context**
None

theeldermillenial:
When either a `PlutusV1Script` or a `PlutusV2Script` is added as a script input and serialized to cbor, any attempt to deserialize throws a deserialization error. This is because there is no explicit catch in `_restore_typed_primitive` for `bytes` data to parse into the appropriate data type except for `ByteString`.

This PR takes a conservative approach to resolving this issue, by only looking for classes named `PlutusVxScript` with a bytes values and casts them to the appropriate script type.

A more general approach would be to detect if the parsing object is a class and the value is `bytes` and using the abstracted parsing object to wrap the bytes object, and do this for all `bytes` values. However, that was not implemented here because it wasn't clear if this could have bad side effects.

I was able to develop a minimal test example that failed before the fix and passed after the fix was implemented.

fixes #251


nielstron (approved):
Looks good to me! Thank you for this fix!

PR #294, opened 2024-01-10
Add support for asset fingerprint (CIP14)

### LINKED ISSUE #289: Support for CIP14
PyCardano does not currently support CIP14, which defines the mechanism for generating assets fingerprints (bech32).

https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0014/

A user requested this in Discord, and I was able to whip up a quick solution. PyCardano has all the necessary pre-requisites to implement this easily. I am posting this here mostly so it doesn't get lost in Discord history. At some point I should open a PR for this under `pycardano.cip`, but for now here is the quick sample I made up.

```python
from nacl.encoding import RawEncoder
from nacl.hash import blake2b
from pycardano.crypto.bech32 import encode

tests = [
    {
        "policy_id": bytes.fromhex(
            "7eae28af2208be856f7a119668ae52a49b73725e326dc16579dcc373"
        ),
        "asset_name": bytes.fromhex(""),
        "asset_fingerprint": "asset1rjklcrnsdzqp65wjgrg55sy9723kw09mlgvlc3",
    }
]

asset_hash = blake2b(
    tests[0]["policy_id"] + tests[0]["asset_name"],
    digest_size=20,
    encoder=RawEncoder,
)
print(f"Calculated Asset Fingerprint: {encode('asset', asset_hash)}")
print(f"Expected Asset Fingerprint: {tests[0]['asset_fingerprint']}")
```

I think to finish this off, it should be:
1. wrapped in a function that integrates well with other PyCardano types (i.e. `MultiAsset` encode/decode similar to `Address` with bech32 addresses)
2. grab all test cases in CIP14 documentation, and use in unit tests

It's not much work, but I'm swamped right now. If someone want's to take a stab at it, be my guest.

theeldermillenial:
This PR adds support for CIP14. I looked into a few different ways to integrate it into multi-assets, and I think there is a way to do it (as suggested in #289) but I figured I would open a PR for the minimal implementation first and see if anyone else thinks its a useful/necessary thing to add.

This also includes unit tests against all examples in the CIP.

For reference:
https://developers.cardano.org/docs/governance/cardano-improvement-proposals/cip-0014/

closes #289


pycardano/cip/cip14.py
```diff
@@ -0,0 +1,33 @@
+from typing import Union
+
+from nacl.encoding import RawEncoder
+from nacl.hash import blake2b
+from pycardano.crypto.bech32 import encode
+
+
+def encode_asset(policy_id: Union[bytes, str], asset_name: Union[bytes, str]) -> str:
```
nielstron: Why not use the ScriptHash and AssetName classes of pycardano here?

theeldermillenial: ... because... Hahaha

I actually had considered that. I decided to just go with raw types. I can switch that over.

Ill add in as an additional input type?

nielstron: Yes I think adding it to the Union would be great :)

theeldermillenial:
Ah, I missed the import and docs. I'll remember that next time. Thanks for the cleanup.

nielstron (approved):
Thanks, looks good to me!

theeldermillenial:
Send it!

PR #297, opened 2024-01-11
WIP: CIP67 and CIP68 Support

theeldermillenial:
This PR is a first pass implementation at handling classes for CIP67 and CIP68. These classes will streamline handling and parsing tokens utilizing these standards. This approach attempts to extend `AssetName` by providing additional validation to ensure CIP68 compliance for token names, as well as associated cbor classes for parsing associated metadata.

Feedback on implementation details and additional functionality is welcomed.

Todo:

- [ ] Unit Tests
- [ ] More rigorous validation of CIP68 fields (e.g. the `uri` fields)
- [ ] Documentation


nielstron:
This is great. I was actually thinking that PlutusData in pycardano could try to support [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) as type annotation for field.

Also I found that deserialization of CIP68 compliant datums fails - have not checked lately if this has been fixed but likely requires a similar fix as #293

theeldermillenial:
> This is great. I was actually thinking that PlutusData in pycardano could try to support [TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) as type annotation for field.

So we just need some kind of extension of this?
https://github.com/Python-Cardano/pycardano/blob/3d27786999eb7ee1b979bf8dafa7b487bfdbabff/pycardano/serialization.py#L806

> 
> Also I found that deserialization of CIP68 compliant datums fails - have not checked lately if this has been fixed but likely requires a similar fix as #293

Have any examples? I just did a query of all cip68 reference tokens and found that 99.99% of them are nfts. I'm trying to figure out a good set of test cases for it. I could use the example in the CIP68 docs, but those are extremely sparse. Ideally I could get a range of different examples.

nielstron:
I would rather extend the support of type annotations for fields of PlutusData here: https://github.com/Python-Cardano/pycardano/blob/3d27786999eb7ee1b979bf8dafa7b487bfdbabff/pycardano/plutus.py#L681

Also for the CIP68 datum I meant the deserialization happening when submitting a transaction with a CIP68 Datum attached - you can try this yourself easily. But don't have any examples at hand right now except for https://preprod.cardanoscan.io/transaction/36f79154eff88d73c21caaf9c5e1bedfc1ff641e39f7f80381dbdba0c104bc50

Cat-Treat:
I'll be working on finishing this off.

Todo:
- Resolve merge conflicts
- Unit tests
- Example of CIP68 token mints

theeldermillenial:
@cffls We are close to having something ready. Just need to get builds passing.

cffls (approved):

PR #300, opened 2024-02-04
Support to/from json raw plutus data

nielstron:
This adds the much requested feature to load and dump unstructured PlutusData (specifically data where the structure is not known to the user). The class "RawPlutusData" is used for this purpose. Note that this also fixes an issue where for example the ScriptContext object can not be dumped becaused it contains dictionaries of the type Dict[bytes, Datum] which contain RawPlutusData.


cffls (approved):
This is amazing! LGTM.

PR #301, opened 2024-02-04
Fix the printing of RawPlutusData

nielstron:
Some recent change in the codebase led to pretty printing of Raw Plutus Data to invoke a pretty printer. This is problematic for two reasons:

- OpShin mimiced the behaviour of PyCardano to print RawPlutusData in the most precise way possible
- The pretty printer re-orders dictionaries (i.e. sorts them). This may lead to the representation of the dictionary not being faithful to the original representation anymore (specifically, could lead to a different CBOR hex)

Another way to tackle this might be to change the behaviour of the pretty printer (inherited from CBORSerializable) to not sort entries in dictionaries anymore (as this could also break CBOR matching)

### DIFF
--- a/pycardano/plutus.py
+++ b/pycardano/plutus.py
@@ -773,6 +773,9 @@ class RawPlutusData(CBORSerializable):
     def __deepcopy__(self, memo):
         return self.__class__.from_cbor(self.to_cbor_hex())
 
+    def __repr__(self):
+        return f"RawPlutusData(data={repr(self.data)})"
+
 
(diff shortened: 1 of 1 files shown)


cffls (approved):

cffls:
You can also try removing this decorator from the class definition:

```python
# @dataclass(repr=False) <------ this line forces it to the repr from parent class, which uses pretty print.
class RawPlutusData(CBORSerializable):
```

nielstron:
Yes that did the trick! Thanks

PR #306, opened 2024-02-09
Accept empty values for restoration

nielstron:
I am not sure why this is forbidden in the first place but it leads to failing deserializations of transactions with just lovelace and not native tokens attached (potentially just produced by cardano-cli > 8.x)


test/pycardano/test_serialization.py
```diff
@@ -301,3 +303,12 @@ class Test(MapCBORSerializable):
...
...264646600200200444a66602c002297ae0132325333015323253330173370e9001000899b8f013375c6038602a0042940c054004c038c04cc038c04c0084cc064008cc0100100044cc010010004c068008c060004dd618009807180498070051180a980b180b00099b8700148004dd6980900098090011bae30100013010002375c601c002646464a66601c602200420022c6eb0c03c004c8c8cc004004008894ccc03c00452f5c0264666444646600200200644a66602a00220062646602e6e9ccc05cdd48031980b9ba9375c60280026602e6ea0dd6980a800a5eb80cc00c00cc064008c05c004dd718070009bab300f001330030033013002301100132323300100100222533300f00114bd6f7b630099191919299980819b8f489000021003133014337606ea4008dd3000998030030019bab3011003375c601e004602600460220026eacc038c03cc03cc03cc03cc01c004c004c0180088c03400452613656375c0024600a6ea80048c00cdd5000ab9a5573aaae7955cfaba05742ae8930011e581c44942234591eee784947f20ab51826428bac8589b372aa9086ee49ee0001825839002b630067f4997129c6fcaea188d4535b1e69fa5860382ba1d2889c3ff3c5cf4a64da3d8d3ac295c3bcbe125961078cdb83080c2a722ef03b1b000000024ab36c59021a0004c829a1008182582064e1867f2c8e8da2e7f5bc06c2224dfdb51b9a40d895d556045ba696483c69685840525f47c570dcfee996eb41a726d1a25e19f46730dccc8e3c05b7dbcf98ceede1ce7931e3b7e0eee9f06b15351042c7285412ac8e4e624ee71805d1a926c65f02f5f6"
```
cffls, line 313: Would it be better to rename this test case to something that contains empty value? It would be also good to assert the target field is actually decoded as an empty value.

cffls (commented):
LGTM. A minor comment on the test.

test/pycardano/test_serialization.py
nielstron: yes thanks for the suggestion! changed this in the latest comment

cffls (approved):

PR #308, opened 2024-02-13
Enable adding scripts with redeemers for withdrawals

nielstron:
This has a usecase in the "withdrawal trick", where only one withdrawal is executed per transaction, checking overall invariants, and all inputs/mints by the script just check that the withdrawal is present.

### DIFF
--- a/pycardano/txbuilder.py
+++ b/pycardano/txbuilder.py
@@ -138,6 +138,10 @@ class TransactionBuilder:
         init=False, default_factory=lambda: []
     )
 
+    _withdrawal_script_to_redeemers: List[
+        Tuple[ScriptType, Optional[Redeemer]]
+    ] = field(init=False, default_factory=lambda: [])
+
     _inputs_to_scripts: Dict[UTxO, ScriptType] = field(
         init=False, default_factory=lambda: {}
     )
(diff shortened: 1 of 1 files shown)


nielstron:
Note that ogmios v6 returns "withdraw:n" as the purpose for a withdrawal, so I changed the tag from "reward" to "withdraw". Not sure if ogmios v5 behaves differently.

Edit: removed from this  PR

pycardano/plutus.py
```diff
@@ -925,7 +925,7 @@ class RedeemerTag(CBORSerializable, Enum):
     SPEND = 0
     MINT = 1
     CERT = 2
-    REWARD = 3
+    WITHDRAW = 3
```
cffls: Renaming this would be a breaking change, which requires major version bump. I would keep using the old name.

nielstron: Moving this to another branch (also not sure if this is an ogmios > 6.x change)

nielstron:
I have confirmed that the `REWARD` tag is actually called `WITHDRAWAL` in Ogmios 5 and `WITHDRAW` in Ogmios 6. The current naming would thus break both the chain contexts either way. Blockfrost also returns `WITHDRAWAL`.

nielstron:
Moreover there is only one place where the redeemer tag is actually used... which is within the transaction builder. So changing the value is most likely not breaking anything relevant. Anyways, I am open to move the change to a seperate PR

cffls:
Thanks for checking! I am fine with the name changing in future PR.

cffls (approved):

pycardano/txbuilder.py
```diff
@@ -735,26 +778,31 @@ def _set_redeemer_index(self):
...
             ):
                 self._inputs_to_redeemers[utxo].index = i
-            elif (
-                utxo in self._inputs_to_redeemers
-                and self._inputs_to_redeemers[utxo].tag == RedeemerTag.MINT
-            ):
-                redeemer = self._inputs_to_redeemers[utxo]
-                redeemer.index = sorted_mint_policies.index(
-                    script_hash(self._inputs_to_scripts[utxo])
-                )
```
nielstron, line 752: Note I removed this case because AFAIK it is impossible to enter this part of the branch. Mint redeemers can only be present in mint scripts, not in inputs.

PR #310, opened 2024-02-17
Fix deserialization of certificates

nielstron:
This fixes deserializing transactions that contain a list of certificates (or more specifically, fixes certificate deserialization from cbor)


cffls (approved):

nielstron:
I think this has been scooped by #321

Issue #292, opened 2024-01-09
Fee off by 1


theeldermillenial:
I'm running into a case where building a transaction with a script input is resulting in an incorrect fee calculation, which is off by 1.

Doing some debugging on `pycardano.utils.fee`, it looks like everything is rounded down with a cast to `int`. Would it make sense to add all raw values and round? In some cases, the values sum to a value with 0.5 ADA getting rounded down, and if we just rounded up this would resolve the issue.

theeldermillenial:
Upon further investigation, we may need to sum raw values and use `math.ceil` for final fee when scripts are involved. I've gotten one example of off by two.

This is occurring when trying to submit through Blockfrost.

cffls:
Makes sense to round up (using `math.cell`) in fee calculations.

cffls:
@theeldermillenial  It would be great if you could provide a failing example so I can find where the root cause is and add it to test.

PR #315, opened 2024-02-20
Add a fee buffer to the transaction builder

nielstron:
In the spirit of #292 this adds a field `fee_buffer` to the transaction builder. This extra buffer makes it possible to submit transactions even if the fee computation is off by a bit (for me often 100-1000 lovelace was totally sufficient). Currently, the only other way to influence the fee computation (without completely doing it on ones own) is to manually adjust the code base.


cffls (approved):
Nice idea!

PR #317, opened 2024-02-21
Allow subtracting assets if not existing before 

nielstron:
Negative values were allowed before as well so it is only consequential to allow subtracting from 0. This simplifies edge case handling for tools that simply want to operate on Values. Is there any inherent rule/implementation in the ledger that lead to the previous error situation?


cffls:
IIRC, I added the check because I didn't think net negative asset was possible in a finalized output. I am fine with this change. We should also add a check in txbuilder for negative asset if there isn't one today.

cffls (approved):

Issue #280, opened 2023-11-08
Support for Pool Certificates 


KINGH242:
**Is your feature request related to a problem? Please describe.**
Parsing the cbor hex from a transaction with pool certificates fail.

**Describe the solution you'd like**
Support the ability to work with transactions involving Pool Certificates.

**Describe alternatives you've considered**
None

**Additional context**
I have already begun working on this and have some working code for handling transactions with pool certificates. Once the code is cleaned up, I will submit a pull request.

cffls:
Completed in https://github.com/Python-Cardano/pycardano/pull/321

PR #321, opened 2024-02-26
Pool certificates

KINGH242:
I finally got around to cleaning up and writing some tests for the code to handle pool certificates as mentioned in #280.

This handles `PoolRegistration` and `PoolRetirement` Certificates. They can be added to transactions, and parsing transactions that contain them will now work correctly.

When registering a pool for the first time, the `TransactionBuilder` has a boolean value called `initial_stake_pool_registration` that can be set to true to add the pool deposit to the transaction.

I also added another boolean value called `witness_override` to the `TransactionBuilder`, which overrides the witness count when estimating the transaction fee. This is very useful for building transactions that multiple parties will sign and witness later.


cffls:
Thanks a ton for adding support for pool certificates! The changes look good overall. There are a few test cases failed when using python3 versions less than `3.10`. I will look into it.

cffls (approved):
Made a few places simpler. LGTM now!

PR #322, opened 2024-03-01
Test for correctness of the supplied script

nielstron:
This is to avoid users accidentally adding script inputs with the wrong script and then get cryptic errors from ogmios.


cffls (approved):
LGTM!

PR #325, opened 2024-03-02
Fix/disallow attach inline datum

### LINKED ISSUE #324: Pycardano does not prevent adding datums for script inputs with inline datums
```python
import pycardano
from pycardano import OgmiosChainContext


context = OgmiosChainContext("ws://localhost:8080", network=pycardano.Network.TESTNET)

us = context.utxos("addr_test1wq6cgea6emrswkj6rctwn7cuvwlt9gs3atxlq5ceukw36nqr5fcy5")
relevant_script_utxo = None
relevant_script_hash = pycardano.ScriptHash(bytes.fromhex("5b65f55ada0c9203408379d64f46d2d7657f27818864e3cd46eb52da"))
for u in us:
    if u.output.script and pycardano.script_hash(u.output.script) == relevant_script_hash:
        relevant_script_utxo = u
        break
utxo_to_spend = context.utxo_by_tx_id("ee3a0fe8e725a875c605be8118a88d33282ff46198bd37b9b2b343cddcf6a52b", 0)
txbuilder = pycardano.TransactionBuilder(context)
txbuilder.add_script_input(utxo_to_spend, script=relevant_script_utxo, redeemer=pycardano.Redeemer(0), datum=pycardano.RawCBOR(bytes.fromhex("d8799f581cab8307abfbb6874dc6692b7930f2b85be7a0c6a8eebf503553523745484144412d4641435440486c6f76656c616365581c1b6b0667bbead0feba17027c752f684019385ee896c63191d1f3a5194b4f7263666178546f6b656e1a04c4b4001a002dc6c0ff")))
secret_key = pycardano.PaymentSigningKey.load("payment.skey")
address = pycardano.Address.from_primitive("addr_test1vz4cxpatlwmgwnwxdy4hjv8jhpd70gxx4rht75p42dfrw3g0p6eeg")
txbuilder.add_input_address(address)
txbuilder.required_signers = [secret_key.to_verification_key().hash()]
tx = txbuilder.build_and_sign([secret_key], change_address=address)
context.submit_tx(tx)
```

This produces the following error:
```
pycardano.exception.TransactionFailedException: [{'unspendableDatums': {'nonSpendable': ['075cb4e39f120c3901ae3fb3d6b8b2bc636e07f5c38be3452a6c74b74ed991cb'], 'acceptable': []}}]

```

Meanwhile, removing the datum parameter makes the transaction go through perfectly fine. I assume hence that attaching datums that are present as inline scripts is forbidden by the ledger - so pycardano should check this during transaction building as well to give a meaningful error message.

nielstron:
fixes #324

### DIFF
--- a/pycardano/backend/ogmios.py
+++ b/pycardano/backend/ogmios.py
@@ -359,8 +359,6 @@ class OgmiosChainContext(ChainContext):
                 )
                 if datum_hash and result.get("datum_type", "inline"):
                     datum = self._get_datum_from_kupo(result["datum_hash"])
-                    if datum is not None:
-                        datum_hash = None
 
                 if not result["value"]["assets"]:
                     tx_out = TransactionOutput(
--- a/pycardano/txbuilder.py
+++ b/pycardano/txbuilder.py
@@ -236,7 +236,7 @@ class TransactionBuilder:
                 f"Datum hash in transaction output is {utxo.output.datum_hash}, "
                 f"but actual datum hash from input datum is {datum_hash(datum)}."
             )
-        if datum is not None and utxo.output.datum is not None:
+        if datum is not None and utxo.output.datum_hash is None and utxo.output.datum is not None:
             raise InvalidArgumentException(
                 f"Inline Datum found in transaction output {utxo.input}, "
                 "so attaching a Datum to the transaction input manually is not allowed."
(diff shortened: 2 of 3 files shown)


cffls:
Looks like an integration test [failed](https://github.com/Python-Cardano/pycardano/actions/runs/8123821854/job/22204632603?pr=325#step:5:4570) due to this change. It turns out the datum from utxo is a datum hash, instead of the actual datum. I think we should allow datum to be attached if its hash was attached to an output. Something like this:

```python
if datum is not None and utxo.output.datum_hash is None and utxo.output.datum is not None:
```

cffls (approved):

PR #328, opened 2024-03-05
Correct variable name and Address input in docs example

B3nac:
Hi all,

Just a small modification to the example that makes it runnable. Maybe Address() at one time required the hash() of both keys that doesn't seem to be the case anymore. Cheers!


nielstron (approved):
Looks like this is indeed an artifact of some previous version with stake key and public key. I think it should be fine to make this example properly enterprise-address based.

cffls (approved):

PR #331, opened 2024-03-14
Support decoding list/indefinite list from cbor

### LINKED ISSUE #311: Reconstructing Transactions from CBOR does not preserve the structure
**Describe the bug**
I am currently looking at implementing chain synchronization based on the building blocks provided by pycardano. For this it is crucial that the transaction id is preserved. There are some transactions for which the reconstructed transaction computes to a different hash than the original transaction.

**To Reproduce**
Example transaction: [f2470890db9c93ed3a98329835a442e70a0b040c53b680a5284be9e80e9d6d89](https://preprod.cardanoscan.io/transaction/f2470890db9c93ed3a98329835a442e70a0b040c53b680a5284be9e80e9d6d89)

```python
from pycardano import Transaction

transaction = Transaction.from_cbor(bytes.fromhex("84a40081825820852ec7f8da4556214f45b166c346802dbe644bdbf16cd8245d431ccdd573fa3100018182581d604b03bd62f7e2d36d157620dd25d3960dc073fa71346a05cb29efbbc91b000000023be7fce3021a00029e61048182018200581cefc7915da7275cfcf0b33909f390d5a2e71e9f1ebc9deaeb503e95bba100828258202fe2a46673313473a680cd2d63993bbf1cd22f864d8d7caeca9c0dffa78e2d595840086ffb2c9adef2a03476f986cf09a50aa40416c007a5f89190ad3edaa8befb5ec7f9dc7a5c6d507705982479a0db4370240e08110b0422ad4de117629f6f7e00825820318e90845971166555968276047c4e3da1102d59b702ec8fc82f748f4072556e584036be05721e2dfbbe35957b0097e53eeffc57c295208bebb8a8063add4dd6978989abd286a8ab57cff4f93d938ab1c4a7ff7089f871b8ba79636c6c3d2fc3ac03f5f6"))
print(transaction.id.payload.hex())
assert transaction.id.payload.hex() == "f2470890db9c93ed3a98329835a442e70a0b040c53b680a5284be9e80e9d6d89", "Incorrect hash"
```

**Expected behavior**
The hash should match

**Environment and software version (please complete the following information):**
 - OS: Ubuntu 21
 - Ogmios 6.0.7
 - PyCardano Version f2596d8114b13cd72ba300f019d91c3750ee3877

### LINKED ISSUE #330: Round trip plutus datum serialization/deserialization error
**Describe the bug**
When adding a script input to a transaction in the transaction builder, the datum is improperly serialized.

**To Reproduce**
I will create a small working example later today, but the basic steps are:
1. Create a custom plutus data class
2. Create a datum, print the datum hash
3. Create a transaction with the transaction builder, add a script input
4. Build the transaction and serialize to cbor
5. Deserialize the transaction, generate the hash of the datum in the witness set

Additional context: Generating the hash of the datum before serialization is correct.

**Expected behavior**
Roundtrip serializing and deserializing of a transaction without mutations in the data.

**Environment and software version (please complete the following information):**
 - OS: Ubuntu
 - PyCardano Version 0.10.0 (also tried with latest on Github)

theeldermillenial:
**tl;dr:**
This PR adds support for proper parsing of fixed vs indefinite lists from CBOR.

**Background**
Currently, pycardano uses a custom encoder to embed indefinite lists when encoding classes to cbor, even if the length of the list is less than 31 elements.

This creates an issue when decoding with `cbor2`, because `cbor2` decodes all lists to the `list` class without any indication of whether the list was encoded as a fixed/indefinite list. This loss of information is problematic because hashes are frequently generated off of CBOR, so encoding as a fixed list vs indefinite list generates a different hash even if the content of the datum is the same.

**Implementation Detail**
This PR is unfortunately dependent on a minor fix in `cbor2`. The `CBORDecoder` class doesn't have a custom decoding mechanism to make modifications to how data is decoded. One approach to fixing this issue would be to subclass `CBORDecoder` to intervene when an indefinite list is found. However, the `CBORDecoder` was implemented in such a way that it relies on a global dictionary that references internal method definitions on a class, making subclassing challenging. To resolve this issue a PR was opened to `cbor2` that simply moves the global dictionaries inside the `CBORDecoder` class, which has no impact on any other section of code:
https://github.com/agronholm/cbor2/pull/225

This PR takes advantage of that PR to subclass the `CBORDecoder`and intervene in the case of an indefinite list. When an indefinite list is discovered, it is cast to `IndefiniteList` and returned, while all other lists are returned as a standard `list`. This required some additional changes, specifically a change how data is being decoded on the `pycardano` side.

This should fix a few different Issues, and fixed a previously undiscovered issue related to roundtrip cbor encoding of datums (datum -> cbor ->datum -> cbor).

fixes #311 
fixes #330


theeldermillenial:
I should mention I found a few other things as well.

Most importantly, this should not be merged in its current state. It might be appropriate to convert this to WIP until the `cbor2` fix is merged.

There seemed to be type errors in code I did not touch. It might be related to the changes I made, but it could also be due to me updating packages.

nielstron:
> pycardano/plutus.py:77: error: Incompatible types in assignment (expression has type "list[Any]", target has type "bytes")  [assignment]
Installing missing stub packages:
pycardano/plutus.py:565: error: Return type "CBORTag" of "to_shallow_primitive" incompatible with return type "list[bytes | bytearray | ByteString | str | int | float | Decimal | bool | tuple[Any, ...] | list[Any] | IndefiniteList | dict[Any, Any] | defaultdict[Any, Any] | OrderedDict[Any, Any] | UndefinedType | datetime | Pattern[Any] | CBORSimpleValue | CBORTag | set[Any] | frozenset[Any] | frozendict[Any, Any] | FrozenList[Any] | IndefiniteFrozenList | None]" in supertype "ArrayCBORSerializable"  [override]
pycardano/plutus.py:791: error: Return type "PlutusData | dict[Any, Any] | int | bytes | IndefiniteList | RawCBOR | CBORTag" of "to_primitive" incompatible with return type "bytes | bytearray | ByteString | str | int | float | Decimal | bool | tuple[Any, ...] | list[Any] | IndefiniteList | dict[Any, Any] | defaultdict[Any, Any] | OrderedDict[Any, Any] | UndefinedType | datetime | Pattern[Any] | CBORSimpleValue | CBORTag | set[Any] | frozenset[Any] | frozendict[Any, Any] | FrozenList[Any] | IndefiniteFrozenList | None" in supertype "CBORSerializable"  [override]
pycardano/cip/cip8.py:120: error: Incompatible types in assignment (expression has type "dict[str, Any]", variable has type "str")  [assignment]
pycardano/transaction.py:296: error: Incompatible types in assignment (expression has type "DatumHash | Any", variable has type "CBORTag")  [assignment]
/home/runner/.cache/pypoetry/virtualenvs/pycardano-XSOAH-2F-py3.10/bin/python -m pip install types-cachetools types-requests

There seem to be some type errors in the static analyzer. Either there is a bug or the analyzer needs more/better type hints

theeldermillenial:
Yes, that is exactly what I was talking about. I'm not entirely sure how those type errors got injected into this with my changes.

On the bright side, I do know this works because I've been using it.

I'm trying to get the attention of the cbor2 maintainer to merge the fix.

nielstron:
I guess suddenly the return type of cbor2.encode which was previously untyped came into scope and unveiled a bunch of mismatching types :)

theeldermillenial:
So...it's my fault that I actually typed it...Lesson learned. Never type anything. 

Thanks for working through it.

nielstron:
By the way,  @theeldermillenial did you consider using tag hooks instead of custom encoding? Or are those overwritten by the major type dict?
https://cbor2.readthedocs.io/en/latest/customizing.html#using-the-cbor-tags-for-custom-types

pyproject.toml
```diff
@@ -22,9 +22,9 @@ repository = "https://github.com/Python-Cardano/pycardano"
 license = "MIT"
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = "^3.8"
```
cffls: Does this feature require python version >= `3.8`? I'd like to keep 3.7 as long as possible.

cffls: Never mind, 3.7 was set to end of life last year :) 
I also bumped the min python version of the entire package to `3.8`

cffls:
Revisiting this PR.. @theeldermillenial is it still WIP, or we can prepare for a merge?

theeldermillenial:
@cffls Unfortunately still WIP. The maintainer of cbor2 wants to make sure that the fix I made has feature parity with the C extension. Unfortunately, that is something that will take more time than I currently have.

I am going to be adding someone to my team, and hopefully they can try to tackle this or free up enough of my time to fix it.

PR #345, opened 2024-05-05
Update node version for integration tests

cffls:
This commit made a few changes:
1. Upgrade cardano-node version to v8.x.x from v1.x.x. This will help with testing when entering Chang hardfork.
2. Upgrade ogmios to 6.x.x because of change 1.
3. Replace built-in ogmios chain context with https://gitlab.com/viperscience/ogmios-python because of change 2.
4. Temporarily disable Kupo in integration test because ogmios-python doesn't support kupo yet.


nielstron (approved):
LGTM

Issue #350, opened 2024-05-23
cardano-cli context incorrect handling latest protocol params


KINGH242:
When using the latest versions of the cardano-cli as `ChainContext`, the calculation for `min_lovelace_post_alonzo` returns 0 because of incorrect handling of the protocol parameters.

To reproduce, build a transaction using the latest cardano-cli as ChainContext, check `min_lovelace_post_alonzo ` output.

```
tx_output = TransactionOutput(address, Value(1000000, multi_assets))

minimum_out_utxo = min_lovelace_post_alonzo(tx_output, context)   # This will be 0
```

I have the fix and will submit a Pull request shortly.

PR #351, opened 2024-05-23
Handle utxoCostPerByte vs coinsPerUtxoByte in cardano-cli chain context protocol params

KINGH242:
This is the fix for #350


cffls:
Looks like some integration tests failed

```
>       return (
            constant_overhead + len(tmp_out.to_cbor())
        ) * context.protocol_param.coins_per_utxo_byte
E       TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'

../pycardano/utils.py:183: TypeError
```

cffls (approved):
LGTM, thanks!

PR #358, opened 2024-06-07
Cannot init blockfrost chain context

thaopt:
fixed: https://github.com/Python-Cardano/pycardano/pull/353

This one should be `str` as defined in the function description.
i cannot initialize the blockfrost chain context due to the recent change:

base_url: Optional[str] = None
...
self._base_url = (base_url) ==> base_url
....
 if "mainnet" in self._base_url.value:
            self._network = Network.MAINNET

self._base_url is str then cannot have attr value (maybe enum item i think) so cannot init the context properly

### DIFF
--- a/pycardano/backend/blockfrost.py
+++ b/pycardano/backend/blockfrost.py
@@ -90,7 +90,7 @@ class BlockFrostChainContext(ChainContext):
         )
 
         # Set network value to mainnet if base_url contains "mainnet".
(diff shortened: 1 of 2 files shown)


thaopt:
I am not sure if have any testcase for this recent change?

cffls:
Thanks for fixing this! I will add some tests to it.

cffls (approved):

PR #363, opened 2024-07-18
Added RawPlutusData to PlutusData primitive types

theeldermillenial:
Some contracts are starting to include forward datums, where a field could contain an arbitrary datum that will be passed on to a subsequent address.

This requires the ability to support arbitrary plutusdata.

In Python <= 3.10, it was possible to pass in `typing.Any` to a field, but starting in `Python 3.11` and beyond, it is no longer possible.

This PR adds `RawPlutusData` to the list of available of PlutusData primitives that can be used for a field.

### DIFF
--- a/pycardano/plutus.py
+++ b/pycardano/plutus.py
@@ -548,7 +548,15 @@ class PlutusData(ArrayCBORSerializable):
         return getattr(cls, k)
 
     def __post_init__(self):
-        valid_types = (PlutusData, dict, IndefiniteList, int, ByteString, bytes)
+        valid_types = (
+            RawPlutusData,
+            PlutusData,
+            dict,
+            IndefiniteList,
+            int,
+            ByteString,
+            bytes,
+        )
         for f in fields(self):
             if inspect.isclass(f.type) and not issubclass(f.type, valid_types):
                 raise TypeError(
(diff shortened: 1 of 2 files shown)


cffls:
Thanks for the change. Could you add a unit test (also as an example) where `RawPlutusData` would be used when deserialized?

cffls (approved):
Thanks for adding the test!

nielstron:
It is (and was) actually possible to use `Datum` as type for arbitrary data in OpShin and AFAIK also in PyCardano. Could you check what difference there is to using RawPlutusData?

theeldermillenial:
@nielstron The issue is that parsing did a hard check on types. You could pass in `PlutusData` as a type, but it would throw an error because of constructor id. It used to be that you could use `typing.Any`, and it would just parse whatever was there, including using `RawPlutusData` to parse a datum. That seems to have changed in Python 3.11+.

nielstron:
I am not talking about `PlutusDatum` but about `Datum`

theeldermillenial:
To my knowledge, there is no `Datum` class in pycardano. I don't use OpShin (yet), so I haven't tried doing that.

Even if there were a `Datum` class, it's not defined in the primitive types and theres an explicit type check here:

https://github.com/Python-Cardano/pycardano/blob/3d65c67883bc90c5e3147ca01027f1cf601ee710/pycardano/plutus.py#L560

Sorry if I'm misunderstanding what you're saying.

PR #368, opened 2024-08-21
Add an OgmiosV6 backend based on ogmios-python

nielstron:
This requires bumping the python version to 3.10+, because ogmios-python requires this. I also refactored the chain backend to not have built-in kupo support but just provide it as an extension using kupo-wrappers. This way, both OgmiosV5 and OgmiosV6 directly have support for kupo.

PR #371, opened 2024-09-02
Chang support

nielstron:
Creating this thread to collect discussion about Chang support. Superseded #368 

TODO
- [x] add tests for ogmiosv6backend tx fetching
- [x] add ogmiosv6backend support for plutusv3
- [x] add tests for kupo backend tx fetching
- [x] disallow 0 and non-empty assets / values
- [x] kupo plutus v3 support


cffls:
All TODOs above seems to be completed:

-  add tests for ogmiosv6backend tx fetching -> https://github.com/Python-Cardano/pycardano/blob/chang/integration-test/test/base.py#L25
-  add ogmiosv6backend support for plutusv3 -> https://github.com/Python-Cardano/pycardano/blob/chang/integration-test/test/test_plutus.py#L388
-  add tests for kupo backend tx fetching -> https://github.com/Python-Cardano/pycardano/blob/chang/integration-test/test/test_plutus.py#L448-L451

nielstron:
I see. So the low coverage on these backends is okay? It would be great if the coverage from the continuous integration could be merged with the main coverage 😅

cffls:
Yeah, it is usually okay if integration test has passed, but definitely better if there are unit tests for them :D

cffls:
It might be possible to combine the coverages. Let me look into this..

nielstron:
I tried something. Should work 😅

nielstron:
Looks good now. I was also trying to integrate #372 but either way this should only increase compatabilities and not break anything. so LGTM

kelizarov:
Hey, appreciate a work on the new update. However we've tried to use `chang` branch on our environment and ran into this issue when trying to call `genesis_param` property function in `OgmiosV6ChainContext`. 
We've been using cardano-node 9.1.0 and ogmios 6.6.0.
Here's the stack trace and exception:
```
   File "/usr/local/lib/python3.10/site-packages/pycardano/backend/ogmios_v6.py", line 192, in genesis_param
     self._genesis_param = self._fetch_genesis_param()  # type: ignore[assignment]
   File "/usr/local/lib/python3.10/site-packages/pycardano/backend/ogmios_v6.py", line 209, in _fetch_genesis_param
     return OgmiosGenesisParameters(client, self._query_current_era())
   File "/usr/local/lib/python3.10/site-packages/ogmios/utils.py", line 31, in __init__
     genesis_parameters, _ = client.query_genesis_configuration.execute(era.value)
   File "/usr/local/lib/python3.10/site-packages/ogmios/statequery/QueryGenesisConfiguration.py", line 46, in execute
     self.send(era, id)
   File "/usr/local/lib/python3.10/site-packages/ogmios/statequery/QueryGenesisConfiguration.py", line 58, in send
     pld = om.QueryNetworkGenesisConfiguration(
   File "/usr/local/lib/python3.10/site-packages/ogmios/model/BaseModelNoException.py", line 9, in __init__
     super(BaseModelNoException, __pydantic_self__).__init__(**data)
   File "/usr/local/lib/python3.10/site-packages/pydantic/v1/main.py", line 339, in __init__
     values, fields_set, validation_error = validate_model(__pydantic_self__.__class__, data)
   File "/usr/local/lib/python3.10/site-packages/pydantic/v1/main.py", line 1074, in validate_model
     v_, errors_ = field.validate(value, values, loc=field.alias, cls=cls_)
   File "/usr/local/lib/python3.10/site-packages/pydantic/v1/fields.py", line 881, in validate
     v, errors = self._validate_singleton(v, values, loc, cls)
   File "/usr/local/lib/python3.10/site-packages/pydantic/v1/fields.py", line 1098, in _validate_singleton
     return self._apply_validators(v, values, loc, cls, self.validators)
   File "/usr/local/lib/python3.10/site-packages/pydantic/v1/fields.py", line 1154, in _apply_validators
     v = validator(cls, v, values, self, self.model_config)
   File "/usr/local/lib/python3.10/site-packages/pydantic/v1/class_validators.py", line 337, in <lambda>
     return lambda cls, v, values, field, config: validator(v)
   File "/usr/local/lib/python3.10/site-packages/pydantic/v1/main.py", line 706, in validate
     return value._copy_and_set_values(value.__dict__, value.__fields_set__, deep=deep_copy)
 AttributeError: 'Params9' object has no attribute '__fields_set__'. Did you mean: '__fields__'?
```

cffls:
Could you try reinstalling all the dependencies? There was an update to ogmios-python.

kelizarov:
I've tried using `ogmios-python-1.2.0` that was coming from this branch and still getting the same error.

nielstron:
@wrmarchetto there appears to be a remaining issue with ogmios-python, can you check it?

wrmarchetto:
@kelizarov @nielstron The ogmios-python error has been fixed in release 1.2.1 (and in the main branch)

kelizarov:
@wrmarchetto  Can confirm the error is no longer present. Thank you!

nielstron:
Requesting review so we can proceed on merging this :)

nielstron:
If I see correctly @wrmarchetto already published an ogmios-package without the OgmiosChainBackend, so there is also no further need to coordinate there. However, should clearly communicate the breaking change. Now users need to rewrite backend to either V5 or V6 explicitely. We could also export the V5 version for backwards compatability as "OgmiosChainBackend"? But I would prefer to keep it explicit since V5 will most likely break in (useful) contexts anyways.

nielstron:
Update: This list of changes was presented appearently at some recent board and summarizes a number of changes. While we don't support all of them yet (i.e. v1 reference scripts) we do support the critical ones in this PR
image

nielstron:
And for reference the changes in lucid:
https://github.com/spacebudz/lucid/pull/254

All of this stolen from the helios discord :)

cffls:
> We could also export the V5 version for backwards compatability as "OgmiosChainBackend"? But I would prefer to keep it explicit since V5 will most likely break in (useful) contexts anyways.

Agree that this release will be a major release with breaking changes. I doubt anyone is using V5 for mainnet, since ogmios v5 isn't compatible with chang? Maybe keep `OgmiosChainBackend` as an alias of ogmios V6?

nielstron:
Ok this makes sense since previously the ogmios backend was also called OgmiosChainBackend
So for users to migrate it would be 
- pycardano.OgmiosChainBackend -> pycardano.OgmiosV5ChainBackend
- ogmios.OgmiosChainBackend -> pycardano.OgmiosChainBackend

Sounds reasonable

cffls:
There is one more PR I'd like to merge before finalizing chang branch: https://github.com/Python-Cardano/pycardano/pull/373

cffls (approved):
I think this is in a good state to merge. @nielstron do you have anything to add?

nielstron:
LGTM

theeldermillenial:
As a heads up, I am testing this out tonight. It does not appear as though some wallets (i.e. Eternl) supports redeemers as maps yet. I am getting errors when generating cbor with pycardano and then sending to Eternl for signing.

Yes, I think this is an Eternl problem, but I did want to put this on the map for anyone else potentially running into this.

nielstron:
I experienced something similar with Mesh. But Eternal works with platforms like MinSwap, so they do support some form of SC txs that is accepted on Chang? Are we missing something?

theeldermillenial:
Can confirm that this is the problem.

https://github.com/Emurgo/cardano-serialization-lib/issues/693

Right now it seems CSL only uses array style redeemers. So this could likely affect multiple wallets.

nielstron:
@theeldermillenial do you know how to fix this? Do we need to update some part on our end or wait for a CSL fix?

nielstron:
According to this comment we should still allow Redeemers as array: https://github.com/Emurgo/cardano-serialization-lib/issues/693#issuecomment-2362959020

Maybe add a switch to the txbuilder for this? @cffls can you look into this since you also added the map support in the first place?

theeldermillenial:
I did some looking into it but haven't had a chance to try it out.

I think there's a one line work around. The Redeemer type is fortunately defined as a list or map. The original redeemer type is still defined, the map was added to it.

I think the workaround looks like this:
1. Build tx with txbuilder
2. Use a list comprehension to convert redeemer map to list

Obviously if you're building the tx manually you can just set the redeemers as an array.

An alternative solution would be to add an option to the txbuilder that specifies redeemer format. By default, it can be set to map, but can be changed to list. Should also be an easy thing to implement.

I'll be able to test out in the next few hours. If you have a comment on this approach, please let me know.

nielstron:
This sounds great @theeldermillenial ! Either approach is fine, maybe just add a function "redeemer_map_to_list" and use that when a flag is set in .build? So both manual builders and the automated builders can benefit from this.

theeldermillenial:
I did some looking into it but haven't had a chance to try it out.

I think there's a one line work around. The Redeemer type is fortunately defined as a list or map. The original redeemer type is still defined, the map was added to it.

I think the workaround looks like this:
1. Build tx with txbuilder
2. Use a list comprehension to convert map to list

Obviously if you're building the tx manually you can just set the redeemers as an array.

An alternative solution would be to add an option to the txbuilder that specifies redeemer format. By default, it can be set to map, but can be changed to list. Should also be an easy thing to implement.

I'll be able to test out in the next few hours. If you have a comment on this approach, please let me know.

theeldermillenial:
That's a good idea. I'll try it out.

theeldermillenial:
Alright, heres my code to swap out the redeemer maps to lists:

```python
    if (
        tx.transaction_witness_set.redeemer is not None
        and len(tx.transaction_witness_set.redeemer) > 0
    ):
        tx.transaction_witness_set.redeemer = [
            pycardano.Redeemer.from_primitive(k.to_primitive() + v.to_primitive())
            for k, v in tx.transaction_witness_set.redeemer.items()
        ]
```

The problem I'm having is that Eternl can't seem to read the cbor after this. I tested building the cbor both with maps and lists, dumping to cbor, and importing into Eternl. In both cases, it will pull up the tx, but if you look at the witness set, it's basically full of null values.

I'll keep hacking away at it, but it looks like this might be more complex than I initially expected.

theeldermillenial:
Nix that, this seems to work. I'll submit a PR today. I had something that was out of alignment with the change branch.

cffls:
Txbuilder always keeps a list of redeemers internally, so the conversion might not be needed. I will try and see if there is a simpler solution to this.

cffls:
Pushed a commit for the fix. 
Now the serialization of redeemers could be configured by setting field `use_redeemer_map` in txbuilders.

theeldermillenial:
@cffls @nielstron 

I actually found an additional error that was causing a different issue in Eternl.

When the `vkey_witnesses` is empty, it encodes an empty value in the map. This causes a problem in Eternl (and I assume this might be a CSL issue) where the entire witness set is read incorrectly. Instead, what is needed is a catch to set `vkey_witnesses=None` when `len(vkey_witnesses)==0`.

I use empty signing keys when building the transaction with txbuilder.

In Eternl, when interacting with the extension through our frontend, we still get the error about redeemers using a map instead of an array, but now I can successfully import the same transaction into Eternl and submit it.

I'll pull in the latest changes and test that along with my fix.

It'll be tomorrow though because I've been at this and a few other days and the brain is fried.

theeldermillenial:
Alright, between the fix from Jerry and this, everything seems to be running smooth.

PR #372, opened 2024-09-02
Automatically reduce 0 and empty multiassets

nielstron:
I noticed that the new ledger rules now disallow empty dicts and 0 values in multiassets. It would be good to anyways prevent this by design - this also always makes transactions cheaper. Here is an implementation proposal. I also tried to verify the non-emptiness strictness with an integration test but it seems to fail. We may want to remove it.

### DIFF
--- a/integration-test/test/test_zero_empty_asset.py
+++ b/integration-test/test/test_zero_empty_asset.py
@@ -149,7 +149,9 @@ class TestZeroEmptyAsset(TestBase):
             address,
             Value(
                 min_val,
-                MultiAsset.from_primitive({policy_vkey.hash().payload: {b"MY_NFT_1": 0}}),
+                MultiAsset.from_primitive(
+                    {policy_vkey.hash().payload: {b"MY_NFT_1": 0}}
+                ),
             ),
         )
         builder.add_output(nft_output)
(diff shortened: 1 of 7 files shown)


nielstron:
This PR does also not yet cover the possibility of creating a 0 value by hand (i.e. accidentally by adding stuff together)

nielstron:
@cffls Can you maybe help with that last bit?
I think the fee computation may be thrown off somewhere - are we computing the wrong number of tokens?

```
FAILED test/test_zero_empty_asset.py::TestZeroEmptyAsset::test_submit_zero_and_empty - ogmios.errors.ResponseError: Ogmios responded with error: {'jsonrpc': '2.0', 'method': 'submitTransaction', 'error': {'code': 3122, 'message': "Insufficient fee! The transaction doesn't not contain enough fee to cover the minimum required by the protocol. Note that fee depends on (a) a flat cost fixed by the protocol, (b) the size of the serialized transaction, (c) the budget allocated for Plutus script execution. The field 'data.minimumRequiredFee' indicates the minimum required fee whereas 'data.providedFee' refers to the fee currently supplied with the transaction.", 'data': {'minimumRequiredFee': {'ada': {'lovelace': 189261}}, 'providedFee': {'ada': {'lovelace': 184861}}}}, 'id': None}

nielstron:
Great, thank you @cffls 
Will merge this into `chang` then unless you have further comments

cffls:
Looks good to me! Please feel free to merge.

PR #376, opened 2024-09-20
Fixed transaction imbalance when burning assets from the same policy

xxAVOGADROxx:
### Problem
The logic did not properly handle cases where assets under one policy were burned. This led to incorrect outputs in the `multi_asset` structure, causing transaction outputs to be improperly computed.

### Solution
A fix was applied to ensure proper handling of burning assets under the same policy. 
- Unite test:
   Assets that are burned under a given policy ID (`policy_id_1`) are properly removed from the `multi_asset` map.

### Changes
- In the `_calc_change` method of `txbuilder.py`.

### Additional information
Tested on [preproduction](https://preprod.cexplorer.io/tx/87375103836ea41590a54403e2c5de7dccc56d72acbae7bc58357cd02e6c9a40).

### DIFF
--- a/pycardano/transaction.py
+++ b/pycardano/transaction.py
@@ -107,7 +107,7 @@ class Asset(DictCBORSerializable):
 
     def __iadd__(self, other: Asset) -> Asset:
         new_item = self + other
-        self.update(new_item)
+        self.data = new_item.data
         return self.normalize()
 
     def __sub__(self, other: Asset) -> Asset:
@@ -173,7 +173,7 @@ class MultiAsset(DictCBORSerializable):
 
     def __iadd__(self, other):
         new_item = self + other
-        self.update(new_item)
+        self.data = new_item.data
         return self.normalize()
 
     def __sub__(self, other: MultiAsset) -> MultiAsset:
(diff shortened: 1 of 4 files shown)


cffls:
Amazing! Thanks for the fix. Could you add a unit test to cover the new code being added?

xxAVOGADROxx:
@cffls Please let me know if any changes are needed. Thanks!

nielstron:
I had hoped that my changes to the multi asset/asset class would disallow empty/ zero assets (see the normalization methods). I am wondering if there is an oversight in the implementation that introduces this need for a fix inside the tx builder? Do we need to run normalization after init as well?

nielstron:
In fact @xxAVOGADROxx I reverted your changes to the txbuilder class and the test case you added still passes. Please make sure that the test case fails before your applied change, otherwise it is not clear what exactly is the change introduced.

I have enabled workflows for this PR so you can directly see if any added unit test increases coverage.

xxAVOGADROxx:
Hello @nielstron, I've added the correct unit test (verifying that it solves the issue). Additionally, I realized the logic for filtering empty/zero assets wasn't necessary, so I've removed that part.

cffls:
Thanks @xxAVOGADROxx for adding the test. I realized that this error originated from `__iadd__` in `Asset` and `MultiAsset`. In the previous implementation, `__iadd__` will only update assets whose value has changed, but it won't update if the key is deleted. I pushed a fix to this behavior and we won't need to change txbuilder now :D

cffls (approved):

PR #386, opened 2024-10-13
Fix failure to report blockfrost error when cbor decode fails

kieransimkin:
Sometimes the blockfrost API does not return a result set containing "result" field, so I was getting an exception out of pycardano which didn't actually show the message I was getting back from blockfrost. I've fixed this so you now get back the error from blockfrost in all circumstances. 

For reference, the error I was getting was: 
`pycardano.exception.TransactionFailedException: Namespace(fault=Namespace(code='client', string='Invalid request: Deserialisation failure while decoding serialised transaction. CBOR failed with error: DeserialiseFailure 0 "expected tag".'), reflection=Namespace(id='efc3ec98-219f-43db-9f58-a6e2ed9fbaea'), servicename='ogmios', type='jsonwsp/fault', version='1.0')`

I am still not entirely sure why my cbor is failing to decode, but I'm pretty sure the problem is with my minting script (from Aiken), not pycardano :)

### DIFF
--- a/pycardano/backend/blockfrost.py
+++ b/pycardano/backend/blockfrost.py
@@ -310,8 +310,13 @@ class BlockFrostChainContext(ChainContext):
             cbor = cbor.hex()
         with tempfile.NamedTemporaryFile(delete=False, mode="w") as f:
             f.write(cbor)
-        result = self.api.transaction_evaluate(f.name).result
+        
+        result = self.api.transaction_evaluate(f.name)
         os.remove(f.name)
+        if not hasattr(result, "result"):
+            raise TransactionFailedException(result)
+        else:
+            result = result.result
         return_val = {}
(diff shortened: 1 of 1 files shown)


cffls:
Thanks for the fix. There is a minor issue in code format when running `make qa`:

```
pycardano/backend/blockfrost.py:313:1: W293 blank line contains whitespace
```

Might be easily fixable through `make format`.

nielstron:
Thanks for your contribution. There seems to be a QA issue, please run `make qa` and fix the reported errors (it's just a blank line in this case)
```
poetry run flake8 pycardano
pycardano/backend/blockfrost.py:313:1: W293 blank line contains whitespace
make: *** [qa] Error 1

```

kieransimkin:
whoops, I've fixed this now :)

cffls (approved):

PR #398, opened 2024-11-16
Add path to Ogmios v6 backend

theeldermillenial:
A couple of months ago, I added a `path` argument to Ogmios.

https://gitlab.com/viperscience/ogmios-python/-/merge_requests/13

This PR adds support for the `path` argument in the Ogmios chain context.

For some additional context, the `path` argument allows you to specify a path in the IP/URL of the Ogmios endpoint. We had a specific use case for this where we wanted to serve Ogmios for different networks on `my.url.io/mainnet` and `my.url.io/preprod`.

I don't believe the change has been deployed on Ogmios yet, so this should be merged once that change is released.

### DIFF
--- a/pycardano/backend/ogmios_v6.py
+++ b/pycardano/backend/ogmios_v6.py
@@ -368,6 +368,7 @@ class OgmiosChainContext(OgmiosV6ChainContext):
 def KupoOgmiosV6ChainContext(
     host: str,
     port: int,
+    path: str,
     secure: bool,
     refetch_chain_tip_interval: Optional[float] = None,
     utxo_cache_size: int = 10000,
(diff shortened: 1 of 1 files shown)


pycardano/backend/ogmios_v6.py
```diff
@@ -263,7 +265,7 @@ def _utxo_from_ogmios_result(self, utxo: OgmiosUtxo) -> UTxO:
             # TODO: Need to test with native scripts
             if script["language"] == "plutus:v3":
                 script = PlutusV3Script(bytes.fromhex(script["cbor"]))
-            elif script["language"] == "plutus:v2":
+            if script["language"] == "plutus:v2":
```
nielstron: this should probably not be here

nielstron (changes requested):
looks good overall, but I think an unintended change snuck in

pycardano/backend/ogmios_v6.py
theeldermillenial: It sure did, I'll fix it.

cffls:
LGTM, pushed a minor fix.

nielstron (approved):

PR #399, opened 2024-11-19
add more examples scripts

Godspeed-exe:
I've added a folder under 'examples' with several specific use case scripts that have been useful for me in the past and I thought it would / could be useful for others.

Let me know if you have any questions or additional requests!


cffls:
Thank you for your contribution! I noticed that there are two pdf files in this PR. Do you mind explaining what they are for?

Godspeed-exe:
you are welcome @cffls 

The PDF files are 'temp files' that were used / are download by the "sign document" scripts. I've added them to .gitignore and deleted them from the repo.

Let me know if you have any further questions!

examples/more_examples/09_sign_document_data.py
```diff
@@ -0,0 +1,103 @@
...
+
+def split_into_64chars(string):
+    return [string[i:i+64] for i in range(0, len(string), 64)]
+
+load_dotenv()
+network = os.getenv('network')
+wallet_mnemonic = os.getenv('wallet_mnemonic')
+blockfrost_api_key = os.getenv('blockfrost_api_key')
+
+file_url = "https://archive.podaac.earthdata.nasa.gov/podaac-ops-cumulus-docs/web-misc/opera/atbd/D-108763_Rev_A_OPERA_DSWx_S1_NI_ATBD_20240530_SIGNED.pdf"
```
cffls: Interesting reference of pdf here :D  It doesn't seem to be a small file though. For the demonstration purpose, I think generating a local temp file with some dummy data might be a better approach. What do you think?

Godspeed-exe: It came from a POC / onsite workshop I did with them. I removed it now and generated a local dummy PDF.

Hope this works for you!

cffls: Somehow missed your comment, my bad. Yes, the changes look good to me. Thank you!

cffls (approved):
LGTM, thank you for your contribution!

PR #400, opened 2024-11-22
ExtendedSigningKey from hd wallet: use key type from class

gszabo:
I noticed this small discrepancy when I tried to derive the staking key from a HD wallet and export it to JSON.


cffls (approved):
LGTM, thanks!

PR #401, opened 2024-11-27
Add Certificate types for staking and voting delegation

gszabo:
Adds support for some of the [new certificate types](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/conway.cddl#L111-L127) of the Conway era.


cffls (approved):
Thanks for your contribution. LGTM!

PR #406, opened 2024-12-13
Add unit tests to transaction.py

tacmota:
Increase test coverage for transaction.py

### DIFF
--- a/pycardano/transaction.py
+++ b/pycardano/transaction.py
@@ -43,6 +43,8 @@ from pycardano.serialization import (
 from pycardano.types import typechecked
 from pycardano.witness import TransactionWitnessSet
 
+import traceback
+
 __all__ = [
     "TransactionInput",
     "AssetName",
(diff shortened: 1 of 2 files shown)


pycardano/transaction.py
```diff
@@ -425,11 +429,6 @@ def __post_init__(self):
 
     def validate(self):
         super().validate()
-        if isinstance(self.amount, int) and self.amount < 0:
```
tacmota, line 428: We don't need to check if amount is int because we already convert int to value in post_init

pycardano/transaction.py
```diff
@@ -88,6 +90,8 @@ def normalize(self) -> Asset:
         for k, v in list(self.items()):
             if v == 0:
                 self.pop(k)
+
+        # traceback.print_stack()
```
nielstron: please remove this leftover comment and the corresponding import :)

nielstron:
Looks good, thank you for your contribution! Just a minor comment to address.

pycardano/transaction.py
tacmota: Thank you! Just removed.

cffls (approved):
LGTM, thanks!

PR #407, opened 2024-12-13
Bump up node version and simplify scripts

cffls:


integration-test/run_pool.sh
```diff
@@ -4,41 +4,18 @@ echo "$NETWORK"
...
-    --tx-out "$(cat /code/keys/pool/full.addr)"+450000000000 \
-    --change-address "$(cat /code/tmp_configs/"$NETWORK"/shelley/utxo-keys/payment.addr)" \
-    --out-file tx.raw \
-    --testnet-magic 42 \
-    --certificate-file /code/keys/pool/stake.cert \
-    --certificate-file /code/keys/pool/pool-registration.cert \
-    --certificate-file /code/keys/pool/delegation.cert \
-    --witness-override 3 \
-    --invalid-hereafter 100000000000000 \
-    --alonzo-era
```
cffls, line 26: New cardano-cli (`>= 9.2.0`) changed the interface of transaction building. Instead of building the txn every time, we will submit a pre-built txn.

PR #409, opened 2024-12-18
Add more tests for serialization

tacmota:


cffls (approved):
Fixed some test cases that has failed in older version of python. LGTM, thanks!

Issue #414, opened 2025-01-04
DeserializeException: ['list'] typed value is required for deserialization. Got <class 'tuple'>


nullhashpixel:
**Describe the bug**

deserialization of some real mainnet txs fails:

```
Traceback (most recent call last):
  File "./test_pycardano.py", line 6, in <module>
    tx = Transaction.from_cbor(tx_cbor)
  File "/site-packages/pycardano/serialization.py", line 485, in from_cbor
    return cls.from_primitive(value)
  File "/site-packages/pycardano/serialization.py", line 162, in wrapper
    return func(cls, value)
  File "/site-packages/pycardano/serialization.py", line 681, in from_primitive
    v = _restore_dataclass_field(f, v)
  File "/site-packages/pycardano/serialization.py", line 506, in _restore_dataclass_field
    return _restore_typed_primitive(f.type, v)
  File "/site-packages/pycardano/serialization.py", line 524, in _restore_typed_primitive
    return t.from_primitive(v)
  File "/site-packages/pycardano/serialization.py", line 162, in wrapper
    return func(cls, value)
  File "/site-packages/pycardano/serialization.py", line 794, in from_primitive
    v = _restore_dataclass_field(f, v)
  File "/site-packages/pycardano/serialization.py", line 505, in _restore_dataclass_field
    return f.metadata["object_hook"](v)
  File "/site-packages/pycardano/serialization.py", line 939, in <lambda>
    return lambda vals: [cls.from_primitive(v) for v in vals]
  File "/site-packages/pycardano/serialization.py", line 939, in <listcomp>
    return lambda vals: [cls.from_primitive(v) for v in vals]
  File "/site-packages/pycardano/serialization.py", line 159, in wrapper
    raise DeserializeException(
[...]

PR #415, opened 2025-01-04
fix type check in deserialization of tx witness (#414)

nullhashpixel:


cffls (approved):
LGTM. Thanks for fixing it!

PR #428, opened 2025-03-01
blockfrost Backend _get_script patches

while0x1:
updated startsWith -> startswith, cast script_type[-1] to int for PlutusScript.from_version method


cffls (approved):
LGTM. Thanks for fixing it!

PR #430, opened 2025-03-04
fix: use OrderedSet or List for pool_owners

### LINKED ISSUE #429: Parsing PoolCertificate in Conway fails
**Describe the bug**
Trying to process a transaction with a PoolCertificate created in Conway era fails since pool_owners is a Set instead of List

**To Reproduce**
Create PoolCertificate using cardano-cli in conway era, then try to parse the CBOR hex.

**Expected behavior**
Should be able to parse certificates from all eras

**Additional context**
Incoming pull request with the fix

KINGH242:
Fixes #429

### DIFF
--- a/pycardano/pool_params.py
+++ b/pycardano/pool_params.py
@@ -22,6 +22,7 @@ from pycardano.serialization import (
     ArrayCBORSerializable,
     CBORSerializable,
     limit_primitive_type,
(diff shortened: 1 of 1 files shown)


cffls (approved):
Pushed a small format fix. LGTM, thanks!

PR #431, opened 2025-03-16
Added IndefiniteDecoder for round trip plutusdata serialization

theeldermillenial:
This PR adds support for IndefiniteDecoder, a special CBOR decoder class that returns indefinite lists as an `IndefiniteList` rather than a `list` as a way to support roundtrip, reproducible `PlutusData` serialization.

There are two new unit tests that demonstrate the need for this PR in the serialization tests.

The first is `test_datum_round_trip`, that creates a custom `PlutusData` class, serializes to CBOR and deserializes. Without the `IndefiniteDecoder`, this test fails.

The second test is `test_datum_raw_round_trip`, that creates a custom `Plutusdata` class, serializes to CBOR and deserializes as `RawPlutusData`. This test is currently marked as an expected fail test, since this PR does not resolve this issue. Attempts to resolve this problem causes issues with other round trip serialization tests, and warrants further investigation.

closes #331

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,6 +28,9 @@ jobs:
       - name: Install dependencies
         run: |
           poetry install
+      - name: Ensure pure cbor2 is installed
+        run: |
+          make ensure-pure-cbor2
       - name: Run unit tests
         run: |
           poetry run pytest --doctest-modules --ignore=examples --cov=pycardano --cov-config=.coveragerc --cov-report=xml
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -24,6 +24,9 @@ jobs:
       - name: Install dependencies
         run: |
           poetry install
+      - name: Ensure pure cbor2 is installed
+        run: |
+          make ensure-pure-cbor2
       - name: Lint with flake8
         run: |
           poetry run flake8 pycardano
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ docs/build
 dist
 .mypy_cache
 coverage.xml
+.cbor2_version
 
 # IDE
 .idea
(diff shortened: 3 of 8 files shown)


theeldermillenial:
Fixed QA issues.

Not entirely sure why but sometimes a test for roundtrip transaction serialization fails. It does not appear to be reproducible.

cffls:
Looks like some non-determinism is introduced, which caused the test to fail randomly.

cffls:
Also, it seems that the customized function `decode_array` is never used in test (see code-cov [report](https://app.codecov.io/gh/Python-Cardano/pycardano/pull/431?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=Python-Cardano#7d602f3932f7f1e1f6fe37bc01625b50-R207)).

cffls:
The reason why it failed non-deterministically was that the order of the elements in a set isn't preserved when deserializing, which led to mismatched order when compared with the original data. 
I figured out that this was due to the library was only using the c-extension of cbor2, which isn't customizable at all. This also explained why your code wasn't even covered by any test before. I added a commit that enforces the library to install cbor2 in pure python, and now your function is being used, but some relevant tests are failing, which is expected because we were expecting list but got IndefinitelyList after your change. It would be great if you could help fix the failed tests. Thanks a lot!

theeldermillenial:
@cffls Wait...so does this mean that the c-extension was also causing the issues related to sets?

I'll try to fix the other outstanding issues tonight.

cffls:
> so does this mean that the c-extension was also causing the issues related to sets?

Yes, it is forcing the code to use c-extension and making it impossible to monkey patch any default behavior, e.g. decoding tag.258 into a set, which doesn't preserve the order.

theeldermillenial:
I wonder if that explains some of the other occasional behavior I would have that cause hash errors. I could never reproduce them and they happened infrequently. 

Sigh...

cffls (approved):
LGTM, thanks!

PR #445, opened 2025-05-19
Coin selection fix

tacmota:
Fix coin selection bug for the test in this branch https://github.com/Python-Cardano/pycardano/tree/coin_selection_fix

The selector didn't initialize with the remaining amount before the selector function started.


cffls (approved):
lgtm, thanks!

PR #446, opened 2025-06-03
Fix transaction certificates in Conway and handle latest cardano-cli transaction commands

KINGH242:
This pull request introduces 3 small but crucial fixes. 

### Certificates type in TransactionBody:
* Modified the `certificates` field in the `TransactionBody` class to accept either a `List` or a `NonEmptyOrderedSet` of `Certificate` objects, providing compatibility with the various ways transactions can be generated.

### Cardano-cli Transaction submission:
* Enhanced the `cardano-cli` transaction submit method command by attempting to use the `latest` prefix for compatability with the latest versions with fallback to the original command.

### Python version compatibility:
* Added a new utility function `greater_than_version` in `pycardano/utils.py` to check if the current Python version is greater than or equal to a specified version. This is used to conditionally import and utilize the `member` attribute from the `enum` module in Python 3.13+ to handle FutureWarning.

### DIFF
--- a/pycardano/backend/cardano_cli.py
+++ b/pycardano/backend/cardano_cli.py
@@ -520,22 +520,39 @@ class CardanoCliChainContext(ChainContext):
 
             try:
                 self._run_command(
-                    ["transaction", "submit", "--tx-file", tmp_tx_file.name]
+                    [
+                        "latest",
+                        "transaction",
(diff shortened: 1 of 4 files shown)


cffls (approved):
LGTM, thanks for the fixes!

PR #448, opened 2025-06-07
Add TextEnvelope class for saving and loading Witness and Transaction files

KINGH242:
This pull request introduces a new `TextEnvelope` class to add JSON serialization and deserialization capabilities. It is added to the `Transaction` and `VerificationKeyWitness` classes, while adding new tests to ensure robustness. The `TextEnvelope` class can later be added to the various Certificates to allow them the same save and load functionality.  Below are the key changes grouped by theme:

### Core Enhancements
* Added a new `TextEnvelope` class in `pycardano/serialization.py` to enable JSON-based serialization and deserialization with fields like `type`, `description`, and `cborHex`. This class also includes methods for saving and loading JSON data to and from files.
* Updated `Transaction` and `VerificationKeyWitness` classes to inherit from `TextEnvelope`, enabling them to utilize its serialization features. Added custom `KEY_TYPE` and `DESCRIPTION` properties to these classes. [[1]](diffhunk://#diff-12bd38147360334930c39af677722ef1157fe518aced11fb50844765cc52eae2L688-R689) [[2]](diffhunk://#diff-02b560c75ee9ac1e0ee0d3d0b50cfe0e2e13c84d36ea708f5276df36636392c5L29-R49)

### Utility and Validation Improvements
* Added a method `is_empty` to `TransactionWitnessSet` for checking whether the witness set is empty.
* Introduced equality (`__eq__`) and shallow primitive conversion (`to_shallow_primitive`) methods to `VerificationKeyWitness` to work with TextEnvelope.

### Testing Enhancements
* Added tests for the `save` and `load` methods of `Transaction` and `VerificationKeyWitness` to validate JSON serialization and file handling. [[1]](diffhunk://#diff-b07e83fd0430f8aae5be30aa11465003aee3783c63de8a28fae1aabd44bb26caR247-R272) [[2]](diffhunk://#diff-acb25430224da43f13611c0fed672d8bbfcc8257ff61ebc0c2c955d11114d549R1-R22)
* Expanded test coverage for `DRep` by introducing parameterized tests to handle various input scenarios, including edge cases and invalid inputs.


pycardano/serialization.py
```diff
@@ -1142,3 +1149,98 @@ def from_primitive(
             raise DeserializeException(f"Invalid {cls.__name__} type {values[0]}")
         # Cast using Type[CodedSerializable] instead of cls directly
         return cast(Type[CodedSerializable], super()).from_primitive(values[1:])
+
+
+@dataclass(repr=False)
+class TextEnvelope(CBORSerializable):
+    """A base class for TextEnvelope types that can be saved and loaded as JSON."""
```
cffls: It might be a good idea to combine this directly with CBORSerializable, which would instantly enable all pycardano classes to save to and load from json. I don't think private fields, `key_type` and `description`, are necessary, because they usually don't change between instances. If we want to customize them for certain child class, the child class can always override the method `json_key_type` or `json_description`. Regarding `payload`, it becomes redundant once the logic is merged to CBORSerializable. The default implementation can directly use the class name as the key type, something like below:

```python3
class CBORSerializable:
    ....
    
    
    @property
    def json_key_type(self) -> str:
        return self.__class__.__name__

    @property
    def json_description(self) -> str:
        return self.__class__.__doc__
    
    def to_json(self) -> str:
        return json.dumps(
            {
                "type": self.json_key_type,
                "description": self.json_description,
                "cborHex": self.to_cbor_hex(),
            }
        )
```

Basic usage

Screenshot 2025-06-08 at 8 06 30 AM

Custom key type
Screenshot 2025-06-08 at 8 12 56 AM

cffls (commented):
Thank you for raising this PR! Enabling classes to save to and load from json is amazing. I think it is possible to simplify the implementation by directly adding the json ability to CBORSerialzable. See the detailed comment below. Let me know if that makes sense to you. Thanks!

pycardano/serialization.py
KINGH242: This can work, but one limitation is that the json files generated by the `cardano-cli` for the same object can have different `type` and `description` depending on which Era it was created in. So being able to set the `type` and `description` has some benefits. The implementation for the `Key` class is a bit mode ideal, but the way Keys handle the payload is a bit different than needed for other classes.  I thought about using it and overriding `from_json` and `from_primitive`, but then it would mess with my developer OCD as, for example, a `Transaction` is not a `Key`🤓.

The `Transaction` and `Witness` files generated and saved to JSON by PyCardano, should be able to work with `cardano-cli`. But if the `type` field has the incorrect text, it will fail. So the way that I currently use it is by getting the Era from the `context`, and using it to generate the correct text for the `type`. I don't think the text in the `description` field is too important as I would usually set it to `Generated with PyCardano`, but the `type` field will fail if it is not what the `cardano-cli` expects, which can change over the eras.

cffls: The current implementation is [hardcoding](https://github.com/Python-Cardano/pycardano/pull/448/files#diff-12bd38147360334930c39af677722ef1157fe518aced11fb50844765cc52eae2R731) the Era as well. 

If the key type and description are dynamic, we can add them `to_json` I guess? Something like this:

```python3
    def to_json(self, key_type: str = None, description: str = None) -> str:
        return json.dumps(
            {
                "type": key_type if key_type else self.json_key_type,
                "description": description if description else self.json_description,
                "cborHex": self.to_cbor_hex(),
            }
        )
```

Also in the `Transaction` class, you just need to override one property without adding `TextEnvelope` type:

```python3
class Transaction:
    @property
    def json_key_type(self) -> str:
        return (
            "Unwitnessed Tx ConwayEra"
            if self.transaction_witness_set.is_empty()
            else "Signed Tx ConwayEra"
        )
```

KINGH242: The current implementation does hardcode it for convenience in the current Era, but it can be set as well during initialization or later.

Passing it as a parameter to the `to_json` should work as well. I will work on refactoring the necessary functions and properties from TextEnvelope to CBORSerializable, then override as necessary. This would be a cleaner solution as the save/load functionality would become immediately available to more classes.

pycardano/serialization.py
```diff
@@ -535,6 +537,118 @@ def from_cbor(cls: Type[CBORBase], payload: Union[str, bytes]) -> CBORBase:
...
+        """
+        key_type = kwargs.pop("key_type", self.json_type)
+        description = kwargs.pop("description", self.json_description)
+        return json.dumps(
+            {
+                "type": key_type,
+                "description": description,
+                "cborHex": self.to_cbor_hex(),
+            }
+        )
```
cffls, line 587: I think it might make sense to add `indent=2` in the `dumps` because it would make the json more readable.

cffls (commented):
Looks great! Just one minor comment. 
Apologies to the late response (somehow the updates of this PR got buried in my inbox).

pycardano/serialization.py
KINGH242: No worries. I will add it today.

KINGH242: I added indent and also fixed the witness to remove the type and description when printing the transaction.

pycardano/serialization.py
```diff
@@ -535,6 +537,119 @@ def from_cbor(cls: Type[CBORBase], payload: Union[str, bytes]) -> CBORBase:
...
+    def json_description(self) -> str:
+        """
+        Return the docstring of the CBORSerializable object's class.
+
+        This property provides a default string description of the object for use in JSON serialization.
+
+        Returns:
+            str: The docstring of the object's class.
+        """
+        return "Generated with PyCardano"
```
cffls: Any reason why not using `self.__class__.__doc__` here?

KINGH242: It has been a while now, but if I am not mistaken, using `self.__class__.__doc__` caused errors and/or type issues.

KINGH242: I just tried it again and qa was failing because `self.__class__.__doc__` has type `Optional[str]`, but `json_description` has a return type of `str`.
This is another way that can work.
```return self.__class__.__doc__ or "Generated with PyCardano"```

cffls: `return self.__class__.__doc__ or "Generated with PyCardano"` looks good.

cffls (approved):
Looks great. Thank you for your contribution! I am sure this feature is going to be used a lot.

PR #449, opened 2025-07-24
Support authenticated ogmios

sassythesasquatchh:
Adds support for authenticated Ogmios connections by allowing additional headers to be passed during backend instantiation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ coverage.xml
 # IDE
 .idea
 .code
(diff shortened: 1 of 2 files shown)


cffls:
Thanks for the contribution. I moved `additional_header` to the end of the argument list to make the change max backward compatible with existing code. Also fixed some formatting check failures.

cffls (approved):

Issue #450, opened 2025-08-02
Plutus Data in Transaction can also be NonEmptyOrderedSet


KINGH242:
When working with transactions generated from other libraries, the `plutus_data` in the `transaction` can also be a `NonEmptyOrderedSet`. This can result in an error when PyCardano parses a signed transaction to submit to the blockchain because the `script_data_hash` will not match the computed hash.

https://github.com/IntersectMBO/cardano-ledger/blob/3fa847bd67a6d1c3c2d5960578c993487e9883b0/eras/conway/impl/cddl-files/conway.cddl#L650 

**To Reproduce**
Create a transaction containing `plutus_data` using Eternl, like a swap. Download unsigned transaction, then sign using PyCardano. Submit to blockchain and it will fail with `PPViewHashsesDontMatch`

**Expected behavior**
Should be able to handle transactions with `plutus_data` as a `NonEmptyOrderedSet`.

**Additional context**
I am currently working on a fix. Will submit pull request when done.

cffls:
Closing as the fix is released in `v0.15.0`.

PR #451, opened 2025-08-02
Add support for NonEmptyOrderedSet in Plutus_data

KINGH242:
This pull request improves handling of indefinite lists, update encoding of Plutus data, and expanded test coverage. Key changes include updates to the `OrderedSet` class to support indefinite lists, modifications to the transaction builder and utilities for more robust Plutus data handling, and update tests to validate these updates.

### Enhancements to `OrderedSet` and Serialization:
* Updated the `OrderedSet` class to inherit from `IndefiniteList` and added support for indefinite list detection and representation. This ensures compatibility with CBOR encoding for both definite and indefinite lists (`pycardano/serialization.py`). [[1]](diffhunk://#diff-c4ba70db478a9e6b82d1c340bc3f581282a58b734c4ce68556d01413b8e16a77L1131-R1142) [[2]](diffhunk://#diff-c4ba70db478a9e6b82d1c340bc3f581282a58b734c4ce68556d01413b8e16a77R1152) [[3]](diffhunk://#diff-c4ba70db478a9e6b82d1c340bc3f581282a58b734c4ce68556d01413b8e16a77L1162-R1175)
* Extended `NonEmptyOrderedSet` to accept `IndefiniteList` as input, improving its flexibility (`pycardano/serialization.py`).

### Improvements in Plutus Data Handling:
* Modified `TransactionWitnessSet` to support `NonEmptyOrderedSet` for `plutus_data` (`pycardano/witness.py`). [[1]](diffhunk://#diff-02b560c75ee9ac1e0ee0d3d0b50cfe0e2e13c84d36ea708f5276df36636392c5L117-R112) [[2]](diffhunk://#diff-02b560c75ee9ac1e0ee0d3d0b50cfe0e2e13c84d36ea708f5276df36636392c5R146-R149)
* Updated the `build_witness_set` method to use `NonEmptyOrderedSet` for Plutus data (`pycardano/txbuilder.py`). [[1]](diffhunk://#diff-5f63e63d39d6c07ad21be34383c46afd02929ac237d9af6def0bdaac5dfd663fR1173) [[2]](diffhunk://#diff-5f63e63d39d6c07ad21be34383c46afd02929ac237d9af6def0bdaac5dfd663fR1185-R1187) [[3]](diffhunk://#diff-5f63e63d39d6c07ad21be34383c46afd02929ac237d9af6def0bdaac5dfd663fL1207-R1211)

### Enhancements to Utilities:
* Refactored the `script_data_hash` function to handle `NonEmptyOrderedSet` for datums and ensure canonical CBOR encoding for cost models (`pycardano/utils.py`).

### Test Coverage Expansion:
* Added new test cases to validate `script_data_hash` behavior with `RedeemerMap` and updated expected hash values to reflect changes in encoding logic (`test/pycardano/test_util.py`).
* Adjusted existing tests to align with the updated handling of Plutus data and indefinite lists (`test/pycardano/test_txbuilder.py`).

### DIFF
--- a/pycardano/utils.py
+++ b/pycardano/utils.py
@@ -13,7 +13,7 @@ from nacl.hash import blake2b
 from pycardano.backend.base import ChainContext
 from pycardano.hash import SCRIPT_DATA_HASH_SIZE, SCRIPT_HASH_SIZE, ScriptDataHash
 from pycardano.plutus import COST_MODELS, CostModels, Datum, Redeemers
-from pycardano.serialization import default_encoder
+from pycardano.serialization import NonEmptyOrderedSet, default_encoder
 from pycardano.transaction import MultiAsset, TransactionOutput, Value
 
 __all__ = [
@@ -235,35 +235,54 @@ def min_lovelace_post_alonzo(output: TransactionOutput, context: ChainContext) -
 
 
 def script_data_hash(
-    redeemers: Redeemers,
-    datums: List[Datum],
+    redeemers: Optional[Redeemers] = None,
+    datums: Optional[Union[List[Datum], NonEmptyOrderedSet[Datum]]] = None,
     cost_models: Optional[Union[CostModels, Dict]] = None,
 ) -> ScriptDataHash:
     """Calculate plutus script data hash
 
     Args:
-        redeemers (Redeemers): Redeemers to include.
-        datums (List[Datum]): Datums to include.
+        redeemers (Optional[Redeemers]): Redeemers to include.
+        datums (Optional[Union[List[Datum], NonEmptyOrderedSet[Datum]]]): Datums to include.
         cost_models (Optional[CostModels]): Cost models.
 
     Returns:
         ScriptDataHash: Plutus script data hash
     """
+    # Handle empty redeemers case - should be encoded as an empty map (A0 in hex)
     if not redeemers:
+        redeemer_bytes = cbor2.dumps({}, default=default_encoder)
         cost_models = {}
-    elif not cost_models:
-        cost_models = COST_MODELS
+    else:
+        redeemer_bytes = cbor2.dumps(redeemers, default=default_encoder)
+        if not cost_models:
+            cost_models = COST_MODELS
 
-    redeemer_bytes = cbor2.dumps(redeemers, default=default_encoder)
+    # Handle datums - if no datums, use empty bytestring
     if datums:
-        datum_bytes = cbor2.dumps(datums, default=default_encoder)
+        if isinstance(datums, list):
+            # If datums is a NonEmptyOrderedSet, convert it to a shallow primitive representation
+            # to ensure correct CBOR encoding
+            datums = NonEmptyOrderedSet(datums)
+        datum_bytes = cbor2.dumps(
+            datums.to_shallow_primitive(), default=default_encoder
+        )
     else:
         datum_bytes = b""
-    cost_models_bytes = cbor2.dumps(cost_models, default=default_encoder)
+
+    # Encode cost models - must use definite length encoding
+    cost_models_bytes = cbor2.dumps(
+        cost_models,
+        default=default_encoder,
+        canonical=True,  # Ensures definite length encoding and canonical map keys
+    )
+
+    # Concatenate in order: redeemers || datums || language views
+    data = redeemer_bytes + datum_bytes + cost_models_bytes
 
     return ScriptDataHash(
         blake2b(
-            redeemer_bytes + datum_bytes + cost_models_bytes,
+            data,
             SCRIPT_DATA_HASH_SIZE,
             encoder=RawEncoder,
         )
--- a/pycardano/utils.py
+++ b/pycardano/utils.py
@@ -12,7 +12,7 @@ from nacl.hash import blake2b
 
 from pycardano.backend.base import ChainContext
 from pycardano.hash import SCRIPT_DATA_HASH_SIZE, SCRIPT_HASH_SIZE, ScriptDataHash
-from pycardano.plutus import COST_MODELS, CostModels, Datum, Redeemers
+from pycardano.plutus import COST_MODELS, CostModels, Datum, Redeemers, RedeemerMap
 from pycardano.serialization import NonEmptyOrderedSet, default_encoder
 from pycardano.transaction import MultiAsset, TransactionOutput, Value
 
@@ -249,14 +249,14 @@ def script_data_hash(
     Returns:
         ScriptDataHash: Plutus script data hash
     """
-    # Handle empty redeemers case - should be encoded as an empty map (A0 in hex)
+    # Handle empty redeemers case - should be encoded as an empty RedeemerMap (A0 in hex)
     if not redeemers:
-        redeemer_bytes = cbor2.dumps({}, default=default_encoder)
+        redeemers = RedeemerMap()
         cost_models = {}
-    else:
-        redeemer_bytes = cbor2.dumps(redeemers, default=default_encoder)
-        if not cost_models:
-            cost_models = COST_MODELS
+    elif not cost_models:
+        cost_models = COST_MODELS
+
+    redeemer_bytes = cbor2.dumps(redeemers, default=default_encoder)
 
     # Handle datums - if no datums, use empty bytestring
     if datums:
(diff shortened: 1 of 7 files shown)


KINGH242:
This fixes #450

pycardano/utils.py
```diff
@@ -235,35 +235,54 @@ def min_lovelace_post_alonzo(output: TransactionOutput, context: ChainContext) -
...
+        redeemers (Optional[Redeemers]): Redeemers to include.
+        datums (Optional[Union[List[Datum], NonEmptyOrderedSet[Datum]]]): Datums to include.
         cost_models (Optional[CostModels]): Cost models.
 
     Returns:
         ScriptDataHash: Plutus script data hash
     """
+    # Handle empty redeemers case - should be encoded as an empty map (A0 in hex)
     if not redeemers:
+        redeemer_bytes = cbor2.dumps({}, default=default_encoder)
```
cffls: We should just check if the redeemer is actually None, and if so, assign an empty redeemer instance to it: `redeemers = Redeemers()`, similar to the implementation [here](https://github.com/Python-Cardano/pycardano/blob/30712f48e32d73fd5fe9e4eec5e70131d1623510/pycardano/txbuilder.py#L585.), and the encoding will be automatically handled.

```suggestion
        if redeemers is None:
            redeemers = Redeemers()
```

Empty `Redeemers()` will be encoded as `A0`.

pycardano/utils.py
```diff
@@ -235,35 +235,54 @@ def min_lovelace_post_alonzo(output: TransactionOutput, context: ChainContext) -
...
+            cost_models = COST_MODELS
 
-    redeemer_bytes = cbor2.dumps(redeemers, default=default_encoder)
+    # Handle datums - if no datums, use empty bytestring
     if datums:
-        datum_bytes = cbor2.dumps(datums, default=default_encoder)
+        if isinstance(datums, list):
+            # If datums is a NonEmptyOrderedSet, convert it to a shallow primitive representation
+            # to ensure correct CBOR encoding
+            datums = NonEmptyOrderedSet(datums)
```
cffls: Not sure if it is a good idea to convert it to `NonEmptyOrderedSet`. Ideally, `script_data_hash` should never change its input, because all items, including redeemers and datums, should match exactly to their representation in the witness set. See this: https://github.com/IntersectMBO/cardano-ledger/blob/1f0c6a4eaa4fb8f937c30a4608a4fafedaca216e/eras/conway/impl/cddl-files/conway.cddl#L449-L451

test/pycardano/test_util.py
```diff
@@ -156,22 +166,39 @@ def test_script_data_hash():
     redeemers = [Redeemer(unit, ExecutionUnits(1000000, 1000000))]
     redeemers[0].tag = RedeemerTag.SPEND
     assert ScriptDataHash.from_primitive(
-        "032d812ee0731af78fe4ec67e4d30d16313c09e6fb675af28f825797e8b5621d"
+        "2ad155a692b0ddb6752df485de0a6bdb947757f9f998ff34a6f4b06ca0664fbe"
```
cffls: This seems to violate the purpose of unit test. The test was written to ensure a known working datum hash remains working. If it has to change, it means the code changes have broken something. If the purpose is to test the case where `datums` is wrapped in `NonEmptyOrderedSet`, please create a new test.

test/pycardano/test_util.py
```diff
@@ -156,22 +166,39 @@ def test_script_data_hash():
...
+    assert ScriptDataHash.from_primitive(
+        "04ad5eb241d1ede2bbbd60c5853de7659d2ecfb1a29d6cbb6921ef7bdd46ca3c"
+    ) == script_data_hash(redeemers=redeemers, datums=[unit], cost_models=cost_models)
+
+
 def test_script_data_hash_datum_only():
     unit = Unit()
     assert ScriptDataHash.from_primitive(
-        "2f50ea2546f8ce020ca45bfcf2abeb02ff18af2283466f888ae489184b3d2d39"
+        "264ea21d9904cd72ce5038fa60e0ddd0859383f7fbf60ecec6df22e4c4e34a1f"
```
cffls: Same as above, changing hash in unit test isn't a good idea.

cffls (commented):
Thanks for the PR! I have a basic understanding of what you are trying to accomplish. My main concern was that, with this change, all plutus data will be converted to `NonEmptyOrderedSet`, regardless of their original format. This might break all transactions generated pre-conway. Since you've fixed the serialization part, I don't think the forced conversion is required in `script_data_hash`, but happy to discuss further or help with debugging if you have a concrete example of failure test cases.

KINGH242:
Thanks for the quick feedback. I will work on changes and revert.

pycardano/utils.py
```diff
@@ -235,35 +235,54 @@ def min_lovelace_post_alonzo(output: TransactionOutput, context: ChainContext) -
...
+    # Handle datums - if no datums, use empty bytestring
     if datums:
-        datum_bytes = cbor2.dumps(datums, default=default_encoder)
+        if not isinstance(datums, NonEmptyOrderedSet):
+            # If datums is not a NonEmptyOrderedSet, handle it as a list
+            datum_bytes = cbor2.dumps(datums, default=default_encoder)
+        else:
+            datum_bytes = cbor2.dumps(
+                datums.to_shallow_primitive(), default=default_encoder
+            )
```
cffls: I don't see why this change is needed. The `default_encoder` will be able to call `to_shallow_primitive` when the datum is not a list. We can keep the original implementation.

pycardano/utils.py
```diff
@@ -235,35 +235,54 @@ def min_lovelace_post_alonzo(output: TransactionOutput, context: ChainContext) -
...
     else:
         datum_bytes = b""
-    cost_models_bytes = cbor2.dumps(cost_models, default=default_encoder)
+
+    # Encode cost models - must use definite length encoding
+    cost_models_bytes = cbor2.dumps(
+        cost_models,
+        default=default_encoder,
+        canonical=True,  # Ensures definite length encoding and canonical map keys
```
cffls: This might break existing unit tests.

test/pycardano/test_util.py
```diff
@@ -160,18 +171,37 @@ def test_script_data_hash():
...
+    ) == script_data_hash(
+        redeemers=redeemers, datums=NonEmptyOrderedSet([unit]), cost_models=cost_models
+    )
+
+
 def test_script_data_hash_datum_only():
     unit = Unit()
     assert ScriptDataHash.from_primitive(
-        "2f50ea2546f8ce020ca45bfcf2abeb02ff18af2283466f888ae489184b3d2d39"
+        "244926529564c04ffdea89005076a6b6aac5e4a2f38182cd48bfbc734b3be296"
```
cffls: Again, I don't think we should allow the hash to change in the unit test.

test/pycardano/test_util.py
```diff
@@ -160,18 +171,37 @@ def test_script_data_hash():
...
     ) == script_data_hash(redeemers=[], datums=[unit])
 
 
 def test_script_data_hash_redeemer_only():
     unit = Unit()
     redeemers = []
     assert ScriptDataHash.from_primitive(
-        "a88fe2947b8d45d1f8b798e52174202579ecf847b8f17038c7398103df2d27b0"
+        "9eb0251b2e85b082c3706a3e79b4cf2a2e96f936e912a398591e2486c757f8c1"
```
cffls: Same as above.

pycardano/utils.py
KINGH242: I think this was needed the way it is because NonEmptyOrderedSet also inherits from other types that might cause incorrect behavior here.

KINGH242:
I was able to fix the other issues, thought I reverted the hashes in the previous one but I maybe didn't check them properly. Now `script_data_hash` can handle any of the various scenarios.

pycardano/utils.py
cffls: Thanks for checking. I think there might be some issue in the way the cbor encodes the `OrderedSet`. I will play with it a bit and see if there is a better solution.

cffls: Fixed this issue by removing `list` type from `OrderedSet`. Having `list` type will enable cbor to bypass `to_shallow_primitve()` and directly encode the object as a list.

cffls:
Pushed a more generic fix. @KINGH242  Could you please check if this new patch works for all your use cases?

KINGH242:
Yes, this still works now. I just added one more thing as I discovered a new edge case with another transaction where `plutus_data` is encoded as an `IndefiniteList`.

cffls (approved):
Thanks for the fix!

PR #463, opened 2025-09-27
Remove unncessary requirements for HDWallet derivation

### LINKED ISSUE #460: Allow for non hardened key derivation using BIP32ED25519PublicKey
**Is your feature request related to a problem? Please describe.**
The cardano serialization lib (Javascript) allows for non hardened key derivation using the Bip32PublicKey.
Pycardano only allows HDWallet to derive, and that alone requires a mnemonic or extended private keys.

**Describe the solution you'd like**
It would be nice if non hardened key derivation could be done using the Bip32PublicKey (which should also be able to be instantiated using an extended public key (64-bytes, as used by the cardano serialization lib - not 78, which some libraries use)

**Describe alternatives you've considered**
I couldn't find any other libraries that support this.  I could open up a subprocess and call Node.js, but I'd rather not.

**Additional context**
N/A

cffls:
Fixes https://github.com/Python-Cardano/pycardano/issues/460

### DIFF
--- a/pycardano/crypto/bip32.py
+++ b/pycardano/crypto/bip32.py
@@ -326,7 +326,7 @@ class HDWallet:
             index += 2**31
 
         if private:
-            if not self._xprivate_key:
+            if self._xprivate_key is None:
                 raise ValueError("Missing private key. Can't do private derivation.")
 
             private_node = (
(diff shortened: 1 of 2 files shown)


tacmota (approved):
Looks good to me. Later we can look into the type check problem.

PR #468, opened 2025-10-26
Freeze indefinite lists and definite lists when decoding

nielstron:
When decoding complex Cardano objects, it may happen that internally, there is a map of e.g. PlutusData -> X or Array -> X. This is fine in CBOR, but can lead to issues when translating this into corresponding Python objects which require hashability. I think for this reason, _all_ CBOR lists/maps should be decoded into a frozen variant to avoid errors during translation.

This patch implements the minimal fix to pass a testcase in Mockfrost (test_vesting2.py) (concretely, decoding the script context from cbor may fail here because it contains the datums map which maps arbitrarily complex datums to arbitrary data). However, more instances of decoded dicts and lists are present in serialization and may need to be adapted. I didn't want to do that right away as I may be unaware of things that break when we enforce freezing any decoded datastructure.

### DIFF
--- a/pycardano/serialization.py
+++ b/pycardano/serialization.py
@@ -52,6 +52,7 @@ from pycardano.types import check_type, typechecked
 __all__ = [
     "default_encoder",
     "IndefiniteList",
+    "IndefiniteFrozenList",
     "Primitive",
     "CBORBase",
     "CBORSerializable",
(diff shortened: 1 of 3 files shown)


cffls:
Some test cases failed. Most of them seems to be easily fixable. I will look into it.

cffls:
Fixed the tests. If possible, could you please add the failed test in Mockfrost (test_vesting2.py)? It would help verification.

nielstron:
I'll have to see what is the best way to add them because I think it's occurring when deserializing a complex transaction

nielstron:
@cffls I added test cases that cover all the changed lines. Maybe one can add more test cases to trigger errors for other cbor serializations, but I am not too familiar with all the options there

cffls (approved):
Thanks! The tests look good to me.

PR #470, opened 2025-11-08
Update Plutus gift example to use Plutus V3 from opshin

cffls:
Update docs to use Plutus V3 and simplify a few steps.
Tested the updated guide e2e on preprod.


nielstron (approved):
Thanks a lot! Looks very good now :)

PR #475, opened 2025-12-17
Use cbor2pure by default

cffls:
This commit will ensure [pure python version](https://github.com/cffls/cbor2pure) of cbor2 is used when installing pycardano for the first time, without modifying any dependencies. Users are still free to use cbor2 if speed is preferred over correctness by setting `CBOR_C_EXETENSION=1` in their environment. 

A dynamic cbor2 module is also exported by pycardano, which allows users to use if they want to write custom version of decoder without worrying about using the wrong cbor library with pycardano. 

```
from pycardano.cbor2 import cbor2
```


README.md
```diff
@@ -57,11 +57,13 @@ Install the library using [pip](https://pip.pypa.io/en/stable/):
...
+For some users, the C implementation may not work properly when deserializing cbor data. For example, the order of inputs of a transaction isn't guaranteed to be the same as the order of inputs in the original transaction (details could be found in [this issue](https://github.com/Python-Cardano/pycardano/issues/311)). This would result in a different transaction hash when the transaction is serialized again.
+
+To solve this problem, a fork of cbor2 is created at [cbor2pure](https://github.com/cffls/cbor2pure). This fork removes C extension and only uses pure python for cbor decoding. By default, for correctness, pycardano uses cbor2pure in decoding. However, if speed is preferred over accuracy, users can set `CBOR_C_EXETENSION=1` in their environment, and the default C extension would be used instead.
```
nielstron: Typo: CBOR_C_EXTENSION

Should also be fixed in other places (and possibly in the cbor2pure package)

pycardano/address.py
```diff
@@ -14,10 +14,10 @@
 from typing import Optional, Type, Union
 
 import base58
-import cbor2
 from cbor2 import CBORTag
 from typing_extensions import override
 
+from pycardano import cbor2
```
nielstron: This looks like a source of potentially nasty circular import issues. Maybe move the cbor2 imports to a specific cbor2.py and import as "from pycardano.cbor2 import cbor2"?

nielstron (changes requested):

nielstron:
Thanks for finally tackling this! Overall the approach is sound, just 2 minor notes.

nielstron (approved):

cffls:
Thanks for reviewing and the suggestions! All tests are passing, merging now.

theeldermillenial:
@cffls This is huge! Thanks. It will make things much easier.

Issue #478, opened 2026-02-26
_cbor2.CBORDecodeEOF: premature end of stream


gg0074x:
**Describe the bug**
When using lists on a Datum and signing the transaction i noticed that the program failed because of a decoding issue with lists of certain length

**To Reproduce**
My datum class was:
```py
@dataclass
class Datum(PlutusData):
    CONSTR_ID = 1
    data: list[bytes]
```
And the list contained the bytes for the string "Hello world!" repeated more than 24 times
Signing a transaction with an output with such Datum fails

**Logs**
```sh
Traceback (most recent call last):
  File "/home/arch/send_datum.py", line 33, in <module>
    datum = LargeDatum.from_cbor(utxo.output.datum.cbor)  # ty:ignore[unresolved-attribute]
  File "/home/arch/.venv/lib/python3.14/site-packages/pycardano/serialization.py", line 543, in from_cbor
    value = cbor2.loads(payload)
  File "/home/arch/.venv/lib/python3.14/site-packages/cbor2pure/_decoder.py", line 846, in loads
    ).decode()
      ~~~~~~^^
  File "/home/arch/.venv/lib/python3.14/site-packages/cbor2pure/_decoder.py", line 239, in decode
    return self._decode()
           ~~~~~~~~~~~~^^
  File "/home/arch/.venv/lib/python3.14/site-packages/cbor2pure/_decoder.py", line 226, in _decode
    return decoder(self, subtype)
  File "/home/arch/.venv/lib/python3.14/site-packages/cbor2pure/_decoder.py", line 473, in decode_semantic
    tag.value = self._decode(unshared=True)
                ~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/home/arch/.venv/lib/python3.14/site-packages/cbor2pure/_decoder.py", line 226, in _decode
    return decoder(self, subtype)
  File "/home/arch/.venv/lib/python3.14/site-packages/pycardano/serialization.py", line 203, in decode_array
    ret = IndefiniteFrozenList(list(self.decode_array(subtype=subtype)))
                                    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/arch/.venv/lib/python3.14/site-packages/cbor2pure/_decoder.py", line 411, in decode_array
    value = self._decode()
  File "/home/arch/.venv/lib/python3.14/site-packages/cbor2pure/_decoder.py", line 226, in _decode
    return decoder(self, subtype)
[...]

PR #485, opened 2026-02-28
fix: resolve NamedTemporaryFile permission errors on Windows in tests

Lampard7crypt:


cffls:
https://github.com/Python-Cardano/pycardano/pull/479 should be able to resolve issue #478 . Could you please explain if there is any other issue this PR tries to address?

Lampard7crypt:
> #479 should be able to resolve issue #478 . Could you please explain if there is any other issue this PR tries to address?

1.  decode_arrayinfinite recursion (serialization.py)
PyCardano overrides cbor2's array decoder to preserve indefinite-length CBOR arrays as IndefiniteFrozenList. Both branches of the override called self.decode_array() —which after patching major_decoders[4] re-entered the same function, causing **infinite recursion**. As a side effect, for arrays with 24+ items this double-called _decode_length on the stream, corrupting it and producing CBORDecodeEOF.

Fix: Save major_decoders[4] before patching, then delegate to that saved reference instead of calling self.decode_array(). This breaks the recursion loop entirely.

2. Windows NamedTemporaryFile permission errors (7 tests)
What: On Windows, NamedTemporaryFile() holds an exclusive file lock while open. When save() tried to open the same path for writing, it got PermissionError. Works fine on Linux/macOS where locking is advisory.

Fix: Use delete=False so the lock is released immediately, then manually delete the file with os.unlink() in a finally block.

cffls:
> Fix: Save major_decoders[4] before patching, then delegate to that saved reference instead of calling self.decode_array(). This breaks the recursion loop entirely.

`self.decode_array(subtype)` resolves to CBORDecoder.decode_array, which is the original unpatched instance method on the class — it was never modified. Saving `_original_decode_array = cbor2._decoder.major_decoders[4]` before patching and calling it explicitly does the exact same thing, just with more indirection.

I am fine with the changes for the temp file usage though.

Lampard7crypt:
> > Fix: Save major_decoders[4] before patching, then delegate to that saved reference instead of calling self.decode_array(). This breaks the recursion loop entirely.
> 
> `self.decode_array(subtype)` resolves to CBORDecoder.decode_array, which is the original unpatched instance method on the class — it was never modified. Saving `_original_decode_array = cbor2._decoder.major_decoders[4]` before patching and calling it explicitly does the exact same thing, just with more indirection.
> 
> I am fine with the changes for the temp file usage though.

Oh, thanks for the clarification, you're right that self.decode_array(subtype) resolves to the unpatched class method since we only patched the major_decoders dict and not the class itself, so saving the explicit reference is indeed redundant. I've reverted those changes. The NamedTemporaryFile fixes I've kept as-is.

cffls (approved):
lgtm, thanks!

PR #487, opened 2026-03-02
solved the issue with the pyre configuration and added a pyright conf…

liliankavengi:
…iguration for better type checking. Also updated the txbuilder and certificate modules to improve functionality and added tests for the txbuilder.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,0 +1,3 @@
+version = 1
+revision = 3
+requires-python = ">=3.14"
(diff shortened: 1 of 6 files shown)


uv.lock
```diff
@@ -0,0 +1,3 @@
+version = 1
```
cffls, line 1: I will bring in uv support in a separate PR.

cffls (commented):
Thanks for the PR! Currently, the project is using mypy for static type checking. Not sure if adding a new type checker is worth it. The feature for txbuilder seems useful though. Happy to include it if the checker and uv.lock is removed.

PR #490, opened 2026-05-31
Cip 67 & 68 Support (continued from #297)

SamDelaney:
This builds on the work done in #297 by @theeldermillenial and @Cat-Treat in hopes of closing it out. Happy to have it pulled into that PR if preferred, I just want to confirm the builds pass here.

Changes that are specific to this PR:

- Merged w/ main

- Add crc8 >=v0.1.0 to (confirmed works w/ 0.1.0, 0.2.0, 0.2.1)

- **Fix `IndefiniteList` initialization crash when wrapping `IndefiniteFrozenList`**

`IndefiniteList.__init__` would raise `AttributeError: 'IndefiniteFrozenList' object has no attribute 'data'` when passed an `IndefiniteFrozenList`. `UserList.__init__` detects it as a `UserList` instance and tries to copy via `.data[:]`, but `IndefiniteFrozenList` inherits from `FrozenList` (a C-extension list subclass) which stores items in the list itself, not in a `.data` attribute.

The fix converts `FrozenList` inputs to a plain `list` before delegating to `UserList.__init__`. This surfaces in CIP-68 datum round-trip serialization, where the CBOR decoder produces an `IndefiniteFrozenList` that is then wrapped in a new `IndefiniteList` during `to_shallow_primitive`.


SamDelaney:
Closing in favor of https://github.com/theeldermillenial/pycardano/pull/4

PR #492, opened 2026-06-05
Speed up PlutusData/CBORSerializable decode ~5.4x and encode ~4.3x (no behavior change)

theeldermillenial:
## Summary

A set of **backend-independent** optimizations to pycardano's `(de)serialization` hot paths, primarily benefiting PlutusData- and set-heavy workloads such as on-chain data indexing and transaction building. **No new dependencies and no change to the CBOR library.** Output is byte-for-byte identical (one documented exception noted below). All tests pass on Python 3.9–3.13.

## Motivation

Profiling showed the cost of typed `PlutusData` decode and real-transaction (de)serialization is dominated by repeated, **data-independent** work in pycardano itself — not the CBOR backend: per-node `get_type_hints()` / signature introspection, per-value type dispatch, redundant typeguard return-checks, re-encoding set elements for de-duplication, and defensive `deepcopy`s on every multi-asset encode.

## Changes

**Decode**
- Memoize `get_type_hints(cls)` and `from_primitive` `type_args` introspection per class.
- Cache `PlutusData.__post_init__` field-type validation per class (per-instance byte-length check preserved).
- Resolve each field's decode dispatch once into a memoized **decode plan**, plus per-class array/map field plans, instead of re-running the `issubclass`/`__origin__`/`isinstance` chain on every value.
- `OrderedSet` keys de-duplication by the element's native hash (CBOR-bytes fallback for unhashable elements), avoiding a `dumps()` per element — previously ~78% of real-transaction decode time.
- Guard the speculative Byron-address `cbor2.loads` probe behind a byte-prefix check.

**Encode**
- Route the recursive `to_primitive` descent through an un-annotated worker, and drop the redundant typeguard return-check on `to_validated_primitive`, so the large `Primitive` `Union` isn't re-validated at every node.
- `_dfs`: scalar-leaf fast path and direct `IndefiniteList.data` iteration (avoids the slow `Sequence.__iter__`).
- Cache `dataclasses.fields()` per class in `to_shallow_primitive`.
- Skip `deepcopy().normalize()` in `Asset`/`MultiAsset.to_shallow_primitive` when there are no zero/empty entries.

All caches are keyed by **class, not data** (so the wins hold across millions of *unique* objects) and use `WeakKeyDictionary` so dynamically-created classes are still garbage-collected.

## Benchmarks

Pure-Python CBOR backend, best-of-5 (reproduce with the included `benchmarks/plutus_bench.py`):

| Operation | Before | After | Speedup |
|---|---|---|---|
| Typed `PlutusData` decode (200 fields) | 4942 µs | ~900 µs | **~5.4×** |
| Typed `PlutusData` encode (200 fields) | 13531 µs | ~2285 µs | **~5.9×** |
| Set-heavy transaction decode | 886 µs | 224 µs | **~4×** |
| `MultiAsset.to_shallow_primitive` | 130 µs | 33 µs | **~3.9×** |

## Correctness

- **Decode**: identical behavior — same `DeserializeException` cases, `Union` fallback order, list/dict/`Optional` handling, `IndefiniteList` preservation, `object_hook` metadata, and one-time `f.type` resolution.
- **Encode**: byte-for-byte identical output (verified on transaction / plutus / metadata round-trips).
- **One documented behavior nuance**: for *hashable* `OrderedSet` elements, de-duplication is now by Python `__eq__`/`__hash__` rather than CBOR-byte equality. These coincide for pycardano's set element types (transaction inputs, key hashes, witnesses); *unhashable* elements retain the original CBOR-byte semantics. Covered by new tests.
- Full suite (unit + doctests) passes; `flake8`, `mypy`, `black`, `isort` clean. New code paths are covered by added tests; genuinely-unreachable defensive branches are marked `# pragma: no cover`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pycardano/address.py
+++ b/pycardano/address.py
@@ -418,19 +418,24 @@ class Address(CBORSerializable):
                     raise DecodingException(f"Failed to decode address string: {e}")
 
         # At this point, value is always bytes
-        # Check if it's a Byron address (CBOR with tag 24)
-        try:
-            decoded = cbor2.loads(value)
-            if isinstance(decoded, (tuple, list)) and len(decoded) == 2:
-                if isinstance(decoded[0], CBORTag) and decoded[0].tag == 24:
-                    # This is definitely a Byron address - validate and decode it
-                    return cls._from_byron_cbor(value)
-        except DecodingException:
-            # Byron decoding failed with validation error - re-raise it
-            raise
-        except Exception:
-            # Not Byron CBOR (general CBOR decode error), continue with Shelley decoding
-            pass
+        # Check if it's a Byron address (CBOR with tag 24). A Byron address is a
+        # 2-element CBOR array whose first element is tag 24, i.e. its bytes start with
+        # b"\x82\xd8\x18" (array(2) + tag(24)). Guarding on that prefix avoids a
+        # speculative cbor2.loads() on every (Shelley) address, whose header byte is
+        # never 0x82.
+        if value[:3] == b"\x82\xd8\x18":
+            try:
+                decoded = cbor2.loads(value)
+                if isinstance(decoded, (tuple, list)) and len(decoded) == 2:
+                    if isinstance(decoded[0], CBORTag) and decoded[0].tag == 24:
+                        # This is definitely a Byron address - validate and decode it
+                        return cls._from_byron_cbor(value)
+            except DecodingException:
+                # Byron decoding failed with validation error - re-raise it
+                raise
+            except Exception:
+                # Not Byron CBOR (general CBOR decode error), continue with Shelley
+                pass
 
         # Shelley address decoding (existing logic)
         header = value[0]
--- a/pycardano/address.py
+++ b/pycardano/address.py
@@ -30,6 +30,11 @@ from pycardano.serialization import CBORSerializable, limit_primitive_type
 
 __all__ = ["AddressType", "PointerAddress", "Address"]
 
+# A Byron address is a 2-element CBOR array whose first element is tag 24, so its
+# serialized bytes start with array(2) + tag(24). Guarding on this prefix avoids a
+# speculative ``cbor2.loads`` on every (Shelley) address.
+_BYRON_ADDRESS_CBOR_PREFIX = b"\x82\xd8\x18"
+
 
 class AddressType(Enum):
     """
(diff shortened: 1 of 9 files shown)


pycardano/serialization.py
```diff
@@ -866,18 +1183,13 @@ def from_primitive(
...
             DeserializeException: When the object could not be restored from primitives.
         """
-        all_fields = [f for f in fields(cls) if f.init]
-
-        restored_vals = []
-        type_hints = get_type_hints(cls)
-        for f, v in zip(all_fields, values):
-            if not isclass(f.type):
-                f.type = type_hints[f.name]
-            v = _restore_dataclass_field(f, v)
```
cffls, line 876: Function `_restore_dataclass_field` could be removed since it isn't used anymore.

pycardano/address.py
```diff
@@ -418,19 +418,24 @@ def from_primitive(cls: Type[Address], value: Union[bytes, str]) -> Address:
...
-            raise
-        except Exception:
-            # Not Byron CBOR (general CBOR decode error), continue with Shelley decoding
-            pass
+        # Check if it's a Byron address (CBOR with tag 24). A Byron address is a
+        # 2-element CBOR array whose first element is tag 24, i.e. its bytes start with
+        # b"\x82\xd8\x18" (array(2) + tag(24)). Guarding on that prefix avoids a
+        # speculative cbor2.loads() on every (Shelley) address, whose header byte is
+        # never 0x82.
+        if value[:3] == b"\x82\xd8\x18":
```
cffls: `b"\x82\xd8\x18"` is used in multiple places, including tests. Would be good to make it a constant.

pycardano/serialization.py
```diff
@@ -304,10 +309,38 @@ def to_primitive(self) -> Primitive:
...
+            tv = type(value)
+            # Fast path for scalar leaves (the large majority of nodes), skipping the
+            # isinstance cascade below.
+            if (
+                tv is int
+                or tv is str
+                or tv is bytes
+                or tv is bool
+                or tv is float
+                or value is None
```
cffls: Can we simplify here, something like the following?

```
_SCALAR_TYPES = frozenset({int, str, bytes, bool, float, type(None)})
...
if type(value) in _SCALAR_TYPES:
    return value
```

test/pycardano/test_serialization.py
```diff
@@ -681,6 +681,35 @@ def test_ordered_set_with_complex_types():
...
+    from pycardano.serialization import _restore_typed_primitive
+
+    # Direct entry point: a primitive value passes straight through.
+    assert _restore_typed_primitive(int, 5) == 5
+    # ByteString: a raw bytes value is wrapped; an already-ByteString value passes through.
+    bs = ByteString(b"hello")
+    assert _restore_typed_primitive(ByteString, b"hello") == bs
+    assert _restore_typed_primitive(ByteString, bs) is bs
+
+
```
cffls, line 712: Is it possible to add some tests to ensure the caching continue to work onwards in case future code changes bypass them?

cffls (commented):
Looks pretty good! Added a few minor comments.

Cerkoryn:
We’re interested in this for our defi-kernel trading engine, where reducing repeated datum/transaction serialization work could help performance. Preserving transaction bytes and existing equality/copy behavior is a prerequisite for adoption; we haven’t yet benchmarked this PR against our workload. Tracking the work here rather than maintaining our own optimization fork.

Issue #402, opened 2024-12-05
Unable to recover signature from unsigned_cbor when RawPlutusData in output


cameronalemon:
I am checking that I can sign a transaction using pycardano based on the unsigned cbor from the eternl wallet (which can be downloaded just before signing any transaction). I have successfully done this for a simple ADA transfer to my own wallet, but cannot get the right signature for some DEX transaction.

Everything matches between my attempt to sign the transaction and the equivalent submitted transaction signed using the eternl wallet and then the hash is downloaded via the tx id, except for the signature.

I am using the mnemonic to get the wallet extended signing keys and verification keys using HDWallet.from_mnemonic(mnem)

```
from pycardano import HDWallet, PaymentExtendedVerificationKey, ExtendedSigningKey, BlockFrostChainContext, Network, VerificationKeyWitness, Transaction

blockfrost_api_key = 'mainnet... '
context = BlockFrostChainContext(blockfrost_api_key, network=Network.MAINNET)

mnem = 'attitude inherit ... ' #24 words
wallet = HDWallet.from_mnemonic(mnem)

payment = wallet.derive_from_path("m/1852'/1815'/0'/0/0")
sk = ExtendedSigningKey.from_hdwallet(payment)
vk = PaymentExtendedVerificationKey.from_signing_key(p_skey)

#download the unsigned cbor hex from eternl just before signing
unsigned_cbor_hex = '84a400...'
unsigned_tx = Transaction.from_cbor(unsigned_cbor_hex)

signature = sk.sign(unsigned_tx.transaction_body.hash())
vk_witnesses = [VerificationKeyWitness(vk, signature)]
signed_tx = Transaction(unsigned_tx.transaction_body, TransactionWitnessSet(vkey_witnesses=vk_witnesses), auxiliary_data=unsigned_tx.auxiliary_data)
```

Susankaten:
> I am checking that I can sign a transaction using pycardano based on the unsigned cbor from the eternl wallet (which can be downloaded just before signing any transaction). I have successfully done this for a simple ADA transfer to my own wallet, but cannot get the right signature for some DEX transaction.
> 
> Everything matches between my attempt to sign the transaction and the equivalent submitted transaction signed using the eternl wallet and then the hash is downloaded via the tx id, except for the signature.
> 
> I am using the mnemonic to get the wallet extended signing keys and verification keys using HDWallet.from_mnemonic(mnem)
> 
> ```
[...]

PR #493, opened 2026-06-06
test: regression for #402 — RawPlutusData inline datum CBOR round-trip

OscarOzaine:
## What this does

Adds a regression test that guards against [issue #402](https://github.com/python-cardano/pycardano/issues/402): unsigned DEX transactions exported by wallets (e.g. Eternl) were getting corrupted when PyCardano decoded and re-encoded them.

## Why it matters

These transactions contain outputs with **inline Plutus datums** encoded as indefinite-length CBOR. Before the fixes in #474 and #479, decoding and re-serializing such a transaction changed the body bytes. Since a transaction's id is the blake2b hash of those bytes, the id changed too — which made any externally-computed signature invalid and caused nodes to reject the transaction.

This test makes sure that bug can't quietly come back.

## What the test checks

Using a real Liqwid DEX transaction (Conway era, two outputs with complex inline datums) as a fixture, it verifies:

1. **Byte-exact round-trip** — `from_cbor()` → `to_cbor_hex()` returns the original bytes.
2. **Both inline datums survive** — both outputs still carry their datum after decode.
3. **Stable transaction id** — the recomputed id matches the known-good value.
4. **Signature stays valid** — signing the body hash produces a signature that verifies (checked directly with `nacl`).

## Scope

Test-only change — one new test in `test/pycardano/test_transaction.py`. No library code is touched.

### DIFF
--- a/test/pycardano/test_transaction.py
+++ b/test/pycardano/test_transaction.py
@@ -1,3 +1,4 @@
+import json
 import os
 import tempfile
 from dataclasses import dataclass
@@ -5,6 +6,7 @@ from fractions import Fraction
 from test.pycardano.util import check_two_way_cbor
 
 import pytest
+from nacl.signing import SigningKey as NACLSigningKey
 from typeguard import TypeCheckError
 
 from pycardano import ParameterChangeAction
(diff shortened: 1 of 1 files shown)


cffls:
Thanks for adding the test!

cffls (approved):

PR #494, opened 2026-06-06
security: redact private key material from SigningKey repr/str

OscarOzaine:
## The problem

Calling `repr()` or `str()` on a signing key returned its full JSON, including the **raw private key** in the `cborHex` field. That meant secret keys could leak into:

- log files (anything that logs an object)
- exception tracebacks
- debugger / REPL output

…without the developer ever intending to export the key.

## The fix

- Signing keys (`SigningKey`, `ExtendedSigningKey`, and their subclasses) now print a redacted form instead of their secret material:

  ```
  <PaymentSigningKey type='PaymentSigningKey' hash=a1b2c3d4e5f6a7b8 [REDACTED]>
  ```

  The `hash=` value is a short, **non-reversible** fingerprint derived from the *public* verification-key hash (which is already public on-chain), so it's safe for logs while still letting you tell two keys apart.
- `__str__` now delegates to `__repr__`, so both are redacted.
- The fingerprint helper never raises — a `repr` that throws would break debuggers and logging.

## What does *not* change

Deliberate export paths are untouched: `to_json()`, `to_cbor()`, and `save()` still return/write the real key material. A warning was added to `to_json()` documenting that its output contains secret material. Verification keys are unaffected.

## Tests

New tests in `test/pycardano/test_key.py` confirm `repr`/`str` never contain the private bytes, the fingerprint is stable and public-derived, and the export methods still work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+**Security:**
+
+- `repr()`/`str()` of signing keys no longer expose private key material; use `.to_json()` or `.save()` to export deliberately.
+
 ## [0.8.1] - 2023-04-06
 
 This patch contains a number of bug fixes to `v0.8.0`.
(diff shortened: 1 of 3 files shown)


cffls:
Code looks good. There is a format issue that could be fixed with `make format`.

Cerkoryn:
We use PyCardano in our defi-kernel trading engine. Signing-key material appearing in ordinary repr/str output is an obstacle to safely using the default diagnostics. We’re interested in seeing this redaction land while keeping explicit key export behavior intact.

PR #495, opened 2026-06-06
security: create key files with 0o600 permissions atomically

OscarOzaine:
## The problem

`CBORSerializable.save()` (used to write `.skey` key files) had three weaknesses:

1. **World-readable files** — files were created with the process umask (often `0o644`), so other local users could read your secret keys.
2. **TOCTOU race** — it checked "does the file exist?" and *then* opened it as a separate step, leaving a window an attacker could exploit.
3. **Symlink following** — it would happily write through a pre-existing symlink.

## The fix

`save()` now creates the file **atomically** with `os.open(path, O_WRONLY | O_CREAT | O_EXCL, mode)`:

- `O_EXCL` makes creation atomic — it fails if the path already exists (closing the race) and refuses to follow a pre-existing symlink.
- The file is then `chmod`-ed to the requested mode regardless of umask.
- Default mode is **`0o600`** (owner read/write only).
- On any write failure, the partially-created file is cleaned up.

A new `mode` keyword lets callers of *non-secret* objects opt into wider permissions when they need it.

## Behavior changes to be aware of

- `save()` now **refuses to overwrite any existing file** (previously it only rejected non-empty files — an empty pre-existing file used to be allowed).
- POSIX file modes are advisory on **Windows**; owner-only enforcement is POSIX-only. Windows users should rely on NTFS ACLs or disk encryption. This is now documented in the address guide.

## Tests

All 574 tests pass. Tests were migrated from `NamedTemporaryFile` to pytest's `tmp_path` (so they don't pre-create the target file, which the new `O_EXCL` behavior rejects). New tests cover the `0o600` mode, overwrite refusal, and partial-write cleanup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+**Security:**
+
+- Key files are now created with `0o600` permissions on POSIX and refuse to overwrite or
+  follow existing paths. `CBORSerializable.save()` now writes atomically with
+  `O_EXCL`/`O_CREAT`, closing a TOCTOU race and preventing world-readable key files. A new
+  `mode` keyword allows callers of non-secret objects to opt into wider permissions.
+
 ## [0.8.1] - 2023-04-06
 
 This patch contains a number of bug fixes to `v0.8.0`.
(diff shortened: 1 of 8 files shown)


cffls:
Looks good. There is a conflict and a format issue (could be fixed with `make format`).

Cerkoryn:
We use PyCardano in our defi-kernel trading engine. Safe private-key file creation is an obstacle to relying on the default save path, so we’re interested in seeing this land. The documented overwrite and platform behavior matters for adopting it without surprising existing callers.

PR #496, opened 2026-06-06
security: warn and self-test when CBOR_C_EXTENSION=1

OscarOzaine:
## Background

A transaction's id is the blake2b hash of its body's **CBOR bytes**, so those bytes must be reproducible. The C `cbor2` extension [does not guarantee map/set element order is preserved on decode](https://github.com/agronholm/cbor2/issues/311), so decoding and re-serializing a transaction can change its id — and a node would then reject or misattribute it.

That's why PyCardano defaults to the pure-Python `cbor2pure` backend. But users can still opt into the faster C extension with `CBOR_C_EXTENSION=1`, with no warning that it's unsafe for transaction workflows.

## What this adds

When `CBOR_C_EXTENSION=1` is set, PyCardano now:

1. **Warns at import time** — emits a `RuntimeWarning` and a `logging` warning explaining the hazard and pointing to the README.
2. **Runs a startup self-test** — a tiny "golden" CBOR vector (a tag-258 set holding `[3, 1, 2]` in non-sorted order) is decoded then re-encoded. A correct backend returns the exact same bytes; an order-losing build does not. If the round-trip isn't byte-stable, it raises `PyCardanoException` so the process **fails fast** instead of silently producing wrong transaction ids.

The self-test can be disabled with `PYCARDANO_CBOR_SELFTEST=0` for users who have verified their workflow never decodes-then-reserializes.

## What does *not* change

With the flag unset (the default), behavior is identical — no warnings, no self-test, no API changes.

## Tests

New tests in `test/pycardano/test_cbor.py` cover the warning, the self-test pass/fail paths, and the opt-out env var. README and FAQ are updated.

### DIFF
--- a/pycardano/cbor.py
+++ b/pycardano/cbor.py
@@ -4,11 +4,79 @@ Conditional cbor2 import module.
 This module provides a centralized location for importing cbor2,
 with support for both the C extension (cbor2) and pure Python (cbor2pure) versions.
 Set the environment variable CBOR_C_EXTENSION=1 to use the C extension.
+
+PyCardano anchors transaction identity in CBOR bytes (``tx.id`` is the blake2b
+hash of the transaction body's CBOR). The C ``cbor2`` extension does not
+guarantee preservation of map/set element order on decode (cbor2 issue #311),
+so a decode-then-reserialize flow can change a transaction's id. For that reason
+the pure-Python ``cbor2pure`` backend is the default, and enabling the C
+extension triggers a runtime warning plus a startup round-trip self-test.
 """
 
+import logging
 import os
+import warnings
+
+logger = logging.getLogger("PyCardano")
+
+_C_EXTENSION_WARNING = (
+    "CBOR_C_EXTENSION=1: using the C cbor2 extension. The C extension does not "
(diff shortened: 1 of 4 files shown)


pycardano/cbor.py
```diff
@@ -4,11 +4,79 @@
...
+def _run_cbor_selftest(backend) -> None:
+    """Run a dependency-free decode→encode round-trip check on ``backend``.
+
+    Mirrors PyCardano's runtime configuration by removing the tag-258 semantic
+    decoder before the check (and restoring it afterwards so the rest of the
+    package's setup is unaffected). Raises :class:`PyCardanoException` if the
+    backend is not byte-stable, i.e. if decode→encode would change transaction
+    ids.
+    """
+    from pycardano.exception import PyCardanoException
```
cffls, line 50: Please move this to top

cffls (commented):
There is a format issue, otherwise lgtm

Cerkoryn:
Our defi-kernel trading engine requires byte-exact transaction CBOR round-trips and currently carries decoder configuration plus independent hash checks. The C-decoder ordering hazard is an obstacle to simplifying that setup. We’re interested in this being addressed; our signing checks will remain independent of a startup self-test.

PR #499, opened 2026-07-24
Cap cbor2 below 6.0.0 (fixes issue 498)

pmcorcoran:
cbor2 6.0.0 removed `CBORDecodeValueError` as a backward-incompatible change. 
cbor2pure 5.8.0 still imports that name, so a clean `pip install pycardano` 
resolves cbor2 to 6.1.3 and `import pycardano` raises ImportError immediately.

This isn't visible in CI because poetry.lock pins cbor2 5.8.0 and uv.lock pins 5.9.0,
and CI installs from the lock. Only users installing from PyPI hit it, which is
users following the README.

Lock files are regenerated for the content-hash only. No resolved versions change,
since 5.8.0/5.9.0 already satisfy the new constraint.

This caps the constraint rather than migrating to cbor2 6.x, which is a larger
effort spanning two packages: cbor2pure re-exports `CBORDecodeValueError` (removed
in 6.0) across several modules, and pycardano's own serialization.py imports
`FrozenDict` (renamed to `frozendict` in 6.0) and uses it as a type elsewhere.
Until both are migrated, capping keeps `pip install pycardano` working.


cffls:
Thanks @pmcorcoran ! Looks like the build failed because poetry.lock is outdated, but looks like `poetry lock` should fix it?

PR #500, opened 2026-08-31
Restrict cbor2

nielstron:
Supercedes #499


cffls (approved):

cffls:
This won't be needed if we merge https://github.com/Python-Cardano/pycardano/pull/503 directly, right?

nielstron:
Yes that should be correct

PR #503, opened 2026-08-31
Support cbor2 6

nielstron:
Depends on #502 (`fix/ci-pytest-command`). That PR should be merged first; once it lands in `main`, this PR will contain only the cbor2 6 changes.

## Summary

- require cbor2 6 or newer and Python 3.10 or newer (cbor2 6 requires at least 3.10)
- adapt serialization and CIP-8 decoding to the cbor2 6 public API
- preserve definite and indefinite CBOR array shapes for byte-exact Cardano transaction and datum round trips
- update dependency locks, CI versions, documentation, and tests


Cerkoryn:
CBOR compatibility is an obstacle to simplifying dependencies in our defi-kernel trading engine: we currently pin cbor2 below 6 with a pure-Python build, and Dendrite pins PyCardano 0.18.0. We’re interested in this work landing with byte-exact transaction/datum preservation and a clear supported-version path; we’ll track this thread rather than open a duplicate.

cffls (approved):
Thanks, this is huge! Looks like it is backward incompatible, should be released with a major version bump I guess.

nielstron:
Yes should neither introduce new behavior nor break old behavior.

